PR #16, opened 2025-06-07
Revert https://github.com/fairpm/fair-plugin/pull/5

afragen:
#5 will result in issues for any other users of the Git Updater Lite library and may not solve the issue if another instance of GUL is loaded before FAIR.

I believe the actual solution is to change the REST route on the `api.fair.pm` site so that the GUL library in its unmodified form will function for everyone.

### DIFF
--- a/inc/updater/class-lite.php
+++ b/inc/updater/class-lite.php
@@ -113,7 +113,7 @@ if ( ! class_exists( 'Fragen\\Git_Updater\\Lite' ) ) {
 				return new \WP_Error( 'invalid_domain', 'Invalid update server domain', $this->update_server );
 			}
 			$url      = add_query_arg(
-				array( 'slug', $this->slug ),
+				array( 'slug' => $this->slug ),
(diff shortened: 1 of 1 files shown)


szepeviktor:
This PR adds `wp-json` to the URL.

afragen:
Yes, this PR adds `wp-json` to the REST route for the reasons stated above.

afragen:
Ready for review

Ipstenu (approved):
The code looks fine, but @rmccue should weigh in before we merge

afragen:
FWIW this is the same code as in the current GUL release.

Ipstenu:
> FWIW this is the same code as in the current GUL release.

I know, it's this:

> I believe the actual solution is to change the REST route on the api.fair.pm site so that the GUL library in its unmodified form will function for everyone.

I want Ryan to confirm that's done (or that he's got a ticket for it ;) )

PR #18, opened 2025-06-07
Assets: Fix code comments and add return type

Soean:
Improves the assets function:

- Docs: `replace_emoji_url()` and `replace_emoji_svg_url()` don't have a parameter. 
- The return types are missing. We should add them to be consistent


joedolson:
I've merged @szepeviktor's PR that includes this and several other code improvements; so I'm going to close this one as no longer needed. Thank you for contributing, @Soean!

Issue #17, opened 2025-06-07
Translating the plugin (to Finnish)


TeemuSuoranta:
I'd like to help translating this plugin to my language (Finnish).

Is the proper way for now through PR to this repository directly to languages directory of the plugin?

ghost:
And I'd like to do the same for German.

ideag:
I can contribute a Lithuanian translation.

timiwahalahti:
Hello @TeemuSuoranta, @openzenta and @ideag - thanks for the interest on translating the plugin!

Like you assumed, PR's are more than welcome.

ideag:
Hi @timiwahalahti! 
Just added a PR for `lt_LT`.
I also noticed the plugin does not currently have a `Domain Path: /languages` header, so it is trying to load translations for the global languages dir. 
I assume we do want to load translations from our own package. 
Should I add a separate PR for that, or tack it to the exitsing one?

timiwahalahti:
Whoopsie, good catch 😅 Separate PR would be the way.

ideag:
Done ;)

TeemuSuoranta:
FI submitted in #23

I noticed one localized string is registered wrong so it doesn't show up and I'll sen separate PR about that:

The issue is:
```
__( 'your browser', 'browser version check', 'fair' )
```

That should be:
```
_x( 'your browser', 'browser version check', 'fair' )
```

TeemuSuoranta:
Ah, nevermind, there are more issues. I'll open separate issue

TeemuSuoranta:
Localization fixes PR submitted #24  . All translations need to be supplemented with missing strings after merging (and re-scanning pot).

toineenzo:
I can do Dutch 😍

szepeviktor:
> I noticed one localized string is registered wrong

🇫‍🇮 #32 Errors fixed!

feibisi:
文派开源（WenPai.org）将为 FAIR 项目提供长期中文支持，可以随时下载：

**WenPai.org will provide long-term Chinese language support for the FAIR project. You can download it anytime at:**  

**https://translate.wenpai.org/locale/zh-cn/default/hosting/fair-package-manager/**

Ipstenu:
Closing this - For now, please open a separate PR per language so we can merge them :)

PR #20, opened 2025-06-07
lt_LT translation for FAIR plugin #17

ideag:


timiwahalahti:
Hello and thanks @ideag!

ideag:
Done @timiwahalahti. There are more strings coming with #30 PR, not sure if we want to get those in for this release as well?   We should probably work a string freeze into the release process as well, to give translators time to catch up in the future.

timiwahalahti:
Yeah, there are plans to implement better release scheduling, likely after this release. That will also make the translation process nicer.

timiwahalahti (approved):

PR #23, opened 2025-06-07
fi translation for FAIR plugin #17

TeemuSuoranta:


timiwahalahti:
Hello and thanks @TeemuSuoranta!

PR #24, opened 2025-06-07
Fix localization issues preventing strings from being translated

TeemuSuoranta:
There are issue preventing localization of some strings:

* References to text-domain `git-updater-lite`
* Usage of localization with context that doesn't use `_x()` function

There are two references of strings without text-domain but those exist in core translations so are ok to use


bitfactory-robin-martijn (approved):
LGTM

joedolson:
Thank you, @TeemuSuoranta!

PR #27, opened 2025-06-07
ca translation for FAIR plugin (POT 6-10-25)

marcarmengou:
Catalan translation for FAIR plugin


timiwahalahti:
Hello and thanks @marcarmengou!

marcarmengou:
Thanks @timiwahalahti! I have updated to the latest version of the POT

PR #28, opened 2025-06-07
Add composer.json

philipjohn:
For WordPress projects managed through Composer, the addition of composer.json here allows them to more easily require the plugin.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,30 @@
+{
+    "name": "fairpm/fair-plugin",
+    "description": "Make your site more FAIR",
+	"version": "0.1.0",
+    "type": "wordpress-plugin",
+	"homepage": "https://github.com/fairpm/fair-plugin",
+	"readme": "README.md",
+	"keywords": [
+		"wordpress",
+		"plugin",
+		"FAIR"
+	],
+    "license": "GPLv2",
+    "authors": [
+        {
+            "name": "FAIR Contributors",
+			"email": "operations@fair.pm",
+			"homepage": "https://fair.pm"
+        }
+    ],
+	"support": {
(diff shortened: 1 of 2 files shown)


szepeviktor:
You may want to ignore the lock file too.

composer.json
```diff
@@ -0,0 +1,30 @@
...
+			"homepage": "https://fair.pm"
+		}
+	],
+	"support": {
+		"issues": "https://github.com/fairpm/fair-plugin/issues",
+		"source": "https://github.com/fairpm/fair-plugin/issues",
+		"docs": "https://github.com/fairpm/fair-plugin/blob/main/README.md"
+	},
+	"minimum-stability": "stable",
+	"require": {
```
tomjn, line 27: Shouldn't this also require composer installers? As well as permission to execute that plugin?

composer.json
```diff
@@ -0,0 +1,30 @@
+{
+	"name": "fairpm/fair-plugin",
+	"description": "Make your site more FAIR",
+	"version": "0.1.0",
```
tomjn, line 4: Would this not be handled by github releases and tags? I'm mindful this adds an extra step to release management that could cause issue if it's missed

composer.json
```diff
@@ -0,0 +1,30 @@
+{
+	"name": "fairpm/fair-plugin",
+	"description": "Make your site more FAIR",
```
tomjn, line 3: This should describe what the plugin does

afragen:
A composer.json in #60

costdev:
#60 contained a `composer.json` file and was merged. If there are any updates or improvements to make to that file, feel free to repurpose this PR.

philipjohn:
Ah that other composer.json is much better!

Issue #9, opened 2025-06-06
Ensure that Site Health is not pinging .org


Ipstenu:
Site Health has a check specifically for the dotorg api, it can be overwritten with filters though, there's also a PR on core trac for removing all instances of the dotorg api if a reference is needed, which takes this, and other areas, into consideration.

* https://core.trac.wordpress.org/ticket/62132

Suggested solution:

> In `WP_Debug_Data::get_wp_core()`, it tests dotorg communication with a request to 'https://wordpress.org'. If a WP_Error object is returned, its message is output. Otherwise, a message reports a successful connection to dotorg.
> 
> So if we filter `pre_http_request`, check for `$url === 'https://wordpress.org',` and return a `WP_Error` object, we can make the message report that FAIR has successfully blocked dotorg communication.
> 
> `WP_Site_Health::get_test_dotorg_communication()`is arguably trickier. This performs a request to 'https://api.wordpress.org'. If a `WP_Error` object is returned, the message is output, but the result is also marked critical. It's not ideal, and I'd personally prefer being able to mark it as good with a message confirming FAIR blocked it, but I don't see the path to do that right now unless we filter 'pre_http_request' to return a fake successful response, and also filter gettext to replace the success message - also not ideal to be filtering gettext.

Good news, our codebase [already changes](https://github.com/fairpm/fair-plugin/blob/main/inc/default-repo/namespace.php) the URL `get_wp_core` attempts to reach, so we've handled the successful connection.

Followup:

> The check should rather tell that the site can communicate with the repository in use. Changing the messages could probably be done with `site_status_test_result` filter.
>
> In addition to already mentioned, `WP_Site_Health::detect_plugin_theme_auto_update_issues `needs to be tested and verified that it still works as it should after repo change.

We will not be able to prove a negative however, and confirm there are _no_ connections to .org, as there is code outside of Core than can do that.

ideag:
> The check should rather tell that the site can communicate with the repository in use. Changing the messages could probably be done with `site_status_test_result`

So funny thing, the dotorg connectivity check is one of a few, that do not get filtered with `site_status_test_result` (see screenshot below). I calculate 4 checks like that in total, three are in the screenshot, but there is also one about Authorisation headers.

So we'll probably need a different hook to get to those strings. `gettext` sounds like a candidate, but I do not like to touch that one if I can help it.

[...]

PR #30, opened 2025-06-07
rework dotorg_connectivity site health check #9

ideag:
rework dotorg_connectivity site health check to not reference .org specifically

### DIFF
--- a/inc/site-health/namespace.php
+++ b/inc/site-health/namespace.php
@@ -1,0 +1,32 @@
+<?php
+/**
+ * Implements the plugin settings page.
+ *
+ * @package FAIR
+ */
+
+namespace FAIR\Site_Health;
+
+/**
+ * Bootstrap.
+ */
+function bootstrap() {
+	add_action( 'admin_enqueue_scripts', __NAMESPACE__ . '\\enqueue_media_scripts' );
+}
+
+function enqueue_media_scripts( $hook_suffix ) {
+
+	if ( 'site-health.php' !== $hook_suffix ) {
+		return;
+	}
+
+	wp_enqueue_script( 'fair-site-health', esc_url( plugin_dir_url( \FAIR\PLUGIN_FILE ) . 'assets/js/fair-site-health.js' ), [ 'wp-i18n' ], \FAIR\VERSION, true );
+	wp_localize_script( 'fair-site-health', 'fairSiteHealth',
+		[
+			'defaultRepoDomain' => \FAIR\Default_Repo\get_default_repo_domain(),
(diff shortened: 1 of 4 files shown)


Ipstenu:
I was afraid this would result in having to rebuild the JS. Thank you for tackling this @ideag !

ideag:
Happy to help! 
I'm still not sure what to do with the dotorg url in the `actions` part. It's so generic it seems better to just remove it.
I also hate how I have to regex the error message in, but I see no way around that at the moment.

ideag:
I've updated the regex to support translations as well. Decided to leave the `actions` link to dotorg as is for now. We can revisit if/when we have a different link to use there.

circuitaxel:
Great work discovering the JS approach!

I wonder if we could simplify this, by leveraging the `rest_post_dispatch` filter? That would allow us to combine all the logic in a single PHP filter.

Optionally, filtering the `site_status_tests` and removing the one for `['async']['dotorg_communication']` and replacing it with our own, but then we need to also create the rest lookup and such. It depends entirely on how much control we want over the entire lifecycle of the test.

ideag:
I thought about that, but using a purpose-specific filter seemed cleaner to me. Moving this to `rest_post_dispatch` also does not solve the need to parse the API response with regex, so it doesn't seem to be an improvement in that respect.
Converting this to a non-async check is an option, but a more involved one. I assume dotorg did it this way with performance in mind and I agree with that judgement.

circuitaxel:
Filtering the `site_status_test` doesn't restrict you to only doing synchronous tests, it can be added as a new async test, it does provide the most control, and retains the performance intent behind the original implementation, but does mean we need to write the whole response our selves (but that may be preferable to just conditionally replacing the strings of a response?)

inc/site-health/namespace.php
```diff
@@ -24,9 +24,19 @@ function enqueue_media_scripts( $hook_suffix ) {
...
 		]
 	);
 
 }
 
+function build_error_message_regex() {
+	$regex = str_replace(
+		[ '%1\$s', '%2\$s' ],
+		[ '(?:.*?)', '(.*)'],
+		preg_quote( __( 'Your site is unable to reach WordPress.org at %1$s, and returned the error: %2$s' ) )
```
Ipstenu: :question: Should this still say WordPressorg?

ideag: Yes, this is intentionally using a Core translation string to build the regex which is used to parse out the API error response. It is not used in output.

joedolson:
Closing this, as it's now a total mess to merge. Replaced with https://github.com/fairpm/fair-plugin/pull/337.

PR #31, opened 2025-06-07
Fix a typo in class-lite.php

szepeviktor:
People may stop using FAIR with this misspelling 🙃


Ipstenu (approved):

Ipstenu:
Fo sure. Thank you for catching that.

PR #39, opened 2025-06-08
Fix URL generation in class-lite.php

szepeviktor:
The previous one was unreadable and error prone.


rmccue:
Thanks for the PR! This is an upstream file from @afragen's https://github.com/afragen/git-updater-lite/blob/main/Lite.php if you want to PR there :)

PR #40, opened 2025-06-08
Add Swedish translation (based on POT 2025-06-07)

tobifjellner:


timiwahalahti:
Hello and thanks @tobifjellner!

tobifjellner:
I just updated the translation. But I noted a string where the variables where just formatting of the complete string and should be moved out from the string. When/if that gets merged, we'll have another fuzzy string.

PR #43, opened 2025-06-08
[DOCS] Extend comment to prevent fatal errors on generated strings

ramonfincken:
Extend comment to prevent fatal errors on generated strings


inc/salts/namespace.php
```diff
@@ -8,7 +8,7 @@
 namespace FAIR\Salts;
 
 /**
- * This is the character set used for various functions.
+ * This is the character set used for various functions. Refrain using \ (backslash).
```
jdevalk:
```suggestion
 * This is the character set used for various functions. Refrain from using \ (backslash) to prevent errors.
```

PR #44, opened 2025-06-08
Patch update browsers bin

ramonfincken:
We do not need the line VERSION_CHECK_FOO on ubuntu. Please double check if this is the case for all OS'es.

Also , before running ... you need to actually run npx manually, otherwhise the $() command waits forever to say "yes"
this might be "solved" by doing this with pipe yes ( | yes )


ramonfincken:
update: I did grep on VERSION_CHECK_FOO , no hit

bin/update-browsers.sh
```diff
@@ -3,12 +3,12 @@
 # Update the browser regex in the version check script.
 #
 # Uses a browserslist query to generate this data.
+# Needs you to run `npx browserslist-useragent-regexp "defaults, unreleased versions"` before first use
```
rmccue, line 6: Hmm, why is this required, because npx prompts for install?

ramonfincken: Yes correct. I needed to enter "y"

ramonfincken: so I have npm/npx installed. But the command needed a one time only "y" manual enter to work

rmccue: Got it; instead of this comment, let's add `--yes` to the `npx` command which should ensure it runs automatically instead :)

Ipstenu:
@rmccue Does the code change need to be in this PR? If not, let's see if we can get this in for 1.1

PR #46, opened 2025-06-08
Remove check on stable PHP versions from dashboard widget logic

### LINKED ISSUE #6: Improve logic for Dashboard widget on PHP version
Currently this is what I see. Do we want this?

Image

afragen:
Fixes #6 

Removes check on a `stable` PHP version to allow a pass-through to the remaining logic. The result is that anyone on a stable PHP version will not see the PHP Update Warning dashboard widget.

### DIFF
--- a/inc/version-check/namespace.php
+++ b/inc/version-check/namespace.php
@@ -224,9 +224,9 @@ function check_php_version( string $version ) {
 
 	$cur_branch_data = $branches[ $cur_branch ];
 
-	if ( $cur_branch_data['state'] === 'stable' || $cur_branch_data['state'] === 'security' ) {
+	if ( $cur_branch_data['state'] === 'security' ) {
 		return [
-			// If we're on the stable or secure branches, the recommended version
+			// If we're on the security branches, the recommended version
 			// should be the latest version of this branch.
 			'recommended_version' => $cur_branch_data['latest'],
 			'minimum_version'     => MINIMUM_PHP,


afragen:
Essentially passing the data back to core's logic of what to display.

Ipstenu:
That looks right to me, but I want to defer to Ryan.

inc/version-check/namespace.php
```diff
@@ -224,9 +224,9 @@ function check_php_version( string $version ) {
 
 	$cur_branch_data = $branches[ $cur_branch ];
 
-	if ( $cur_branch_data['state'] === 'stable' || $cur_branch_data['state'] === 'security' ) {
+	if ( $cur_branch_data['state'] === 'security' ) {
 		return [
-			// If we're on the stable or secure branches, the recommended version
+			// If we're on the security branches, the recommended version
```
Ipstenu, line 229: Part of me wants to check for stable, but as an info ... MVP. We can do that kind of thing later with a 'contact your host...'

afragen: From what I see in core, any PHP version listed as `stable` is fine as long as it's on the current minor, not latest.

Ipstenu: Yeah that's why it's a later thing :) We can do better. Like if you're 20 patch releases out of date, warn you.

rmccue (approved):
As discussed in the tech call, this was intentional, but it's also a bit annoying. We should rethink this widget in the future.

PR #49, opened 2025-06-09
it_IT translation for FAIR plugin

andreg:
This pull request adds italian translation for the FAIR plugin.


timiwahalahti:
Hello and thanks @andreg!

We are planning to release the next version on Monday, the 16th, which means a code freeze will be in effect on Friday, the 13th.

_Internal notes: PTE | approved_

andreg:
Hi @timiwahalahti,

I've updated the PR to include the new strings. Will, of course, check again after code freeze. Thank you!

PR #52, opened 2025-06-09
Terminology improvements

milesddavis:
The FAIR project is a constructive one, helping strengthen the future of the whole WordPress ecosystem. The terminology used to describe the plugin should reflect that. Currently the first impression of this plugin from its description is that it's anti-WordPress.org and anti-Automattic, which is not the case.

This PR updates the user-facing readme file so its language and purpose more closely matches what the FAIR project is about and to generally implement a more positive, constructive, forward-looking tone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,24 +6,32 @@ This repository contains the plugin for installation into WordPress.
 
 ## Description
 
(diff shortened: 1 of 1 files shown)


rmccue (approved):
Love this, looks great.

joedolson (approved):
These changes make sense to me. I think they should be accepted as they are; there could be minor changes, but there's certainly nothing blocking.

PR #58, opened 2025-06-09
Keep generate-pot up to date with main

### LINKED ISSUE #56: Fix POT generation action
Noticed after #24 fixed some text domain issues, but PR with those strings didn't come up. 

Action uses `generate-pot` branch which is behind `main`, since if I'm not mistaken the action never updates `generate-pot` with `main` It should do that.

timiwahalahti:
Fixes #56 by merging `main` into `generate-pot` before actual POT generation.

### DIFF
--- a/.github/workflows/generate-pot-pr.yml
+++ b/.github/workflows/generate-pot-pr.yml
@@ -44,7 +44,8 @@ jobs:
         run: |
           git fetch --all --prune
           git checkout generate-pot
-          git pull --no-rebase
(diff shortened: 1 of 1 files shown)


afragen:
It looks good, but I need someone better at the git CLI to approve.

Ipstenu (approved):
I always forget about no edit!

afragen:
This has borked the GA and it no longer works. Perhaps we need another fix. I'm going to revert for now.

PR #63, opened 2025-06-10
Improv comments to translators and removing indirect formatting

tobifjellner:
Added clarification that some printf variables include opening anchors.
Also removed unnecessary formatting from a string, since it was affecting the complete string.

### DIFF
--- a/inc/updater/class-lite.php
+++ b/inc/updater/class-lite.php
@@ -395,7 +395,7 @@ if ( ! class_exists( 'Fragen\\Git_Updater\\Lite' ) ) {
 						);
 					if ( ! empty( $current->response[ $theme->slug ]['package'] ) ) {
 						printf(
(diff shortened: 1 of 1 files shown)


tobifjellner:
Ouch. Scratch this. A new version is coming. Opening anchors and version numbers need to be separate variables in the printf statements, to allow for correct translations.

tobifjellner:
In Swedish, we would typically say "version x.x" or "details of version x.x", i.e. it would be good to NOT have the opening anchor and the version number glued together in one variable. 
But I'm not the right person to make such changes to the code.
So: If there's time to fix that, then this PR should be rejected. In other case, better to add this clarification for now.

joedolson:
I'll take on fixing this; I agree that these strings should be separated.

jdevalk:
I will say that this looks already merge-able as is to me ;)

norcross (approved):

PR #65, opened 2025-06-10
Support ClassicPress News widget

xxsimoxx:
[ClassicPress](https://www.classicpress.net/) is a fork of WordPress, without blocks.

Since ClassicPress have it's own news system, this prevents a fatal error in the dashboard.
`wp_print_community_events_markup()` is not present in ClassicPress as the Community Events widget.

## Before

image

## After

image

### DIFF
--- a/inc/dashboard-widgets/namespace.php
+++ b/inc/dashboard-widgets/namespace.php
@@ -10,7 +10,7 @@ const EVENTS_API = 'https://api.fair.pm/fair/v1/events';
  * Bootstrap.
  */
 function bootstrap() {
-	if ( function_exists( 'classicpress_version' ) ) {
+	if ( ! function_exists( 'wp_print_community_events_markup' ) ) {
 		return;
 	}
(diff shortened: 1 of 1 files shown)


joedolson (changes requested):
I think it would be better to check `! function_exists( 'wp_print_community_events_markup' )`, as that would support any case where this was removed, rather than being specific to Classic Press. Thoughts?

inc/dashboard-widgets/namespace.php
```diff
@@ -10,6 +10,9 @@
  * Bootstrap.
  */
 function bootstrap() {
+	if ( function_exists( 'classicpress_version' ) ) {
```
Ipstenu: `if ( function_exists( 'wp_ajax_get-community-events' ) ) { `

It’s great to check for classic press conflicts, but if we can make our checks universal (in case someone else makes a fork and wants to use this) then it’s better 🙂

Ipstenu (changes requested):

xxsimoxx:
Sure, I agree to the proposed change and just pushed a commit.

PR #71, opened 2025-06-11
Add installer for FAIR protocol

rmccue:
Work in progress. Currently, contains the pieces for installation but not all hooked up properly yet.

To preview the UI, load a DID manually into the plugin info screen; eg `/wp-admin/plugin-install.php?tab=plugin-information&plugin=did%3Aplc%3Aafjf7gsjzsqmgc7dlhb553mv`

### DIFF
--- a/inc/packages/admin/info.php
+++ b/inc/packages/admin/info.php
@@ -1,0 +1,305 @@
+<?php
+
+namespace FAIR\Packages\Admin\Info;
+
+use FAIR\Packages;
+use FAIR\Packages\MetadataDocument;
+use FAIR\Packages\ReleaseDocument;
+
+/**
+ * Sanitize HTML content for plugin information.
+ *
+ * @param string $html The HTML content to sanitize.
+ * @return string Sanitized HTML content.
+ */
+function sanitize_html( string $html ) : string {
+	static $allowed = [
+		'a' => [
+			'href' => [],
+			'title' => [],
+			'target' => [],
+		],
+		'abbr' => [
+			'title' => [],
+		],
+		'acronym' => [
+			'title' => [],
+		],
+		'code' => [],
+		'pre' => [],
+		'em' => [],
+		'strong' => [],
+		'div' => [
+			'class' => [],
+		],
+		'span' => [
+			'class' => [],
+		],
+		'p' => [],
+		'br' => [],
+		'ul' => [],
+		'ol' => [],
+		'li' => [],
+		'h1' => [],
+		'h2' => [],
+		'h3' => [],
+		'h4' => [],
+		'h5' => [],
+		'h6' => [],
+		'img' => [
+			'src' => [],
+			'class' => [],
+			'alt' => [],
+		],
+		'blockquote' => [
+			'cite' => true,
+		],
+	];
+	return wp_kses( $html, $allowed );
+}
+
+function get_section_title( string $id ) {
+	switch ( $id ) {
+		case 'description':
+			return _x( 'Description', 'Plugin installer section title' );
+		case 'installation':
+			return _x( 'Installation', 'Plugin installer section title' );
+		case 'faq':
+			return _x( 'FAQ', 'Plugin installer section title' );
+		case 'screenshots':
+			return _x( 'Screenshots', 'Plugin installer section title' );
+		case 'changelog':
+			return _x( 'Changelog', 'Plugin installer section title' );
+		case 'reviews':
+			return _x( 'Reviews', 'Plugin installer section title' );
+		case 'other_notes':
+			return _x( 'Other Notes', 'Plugin installer section title' );
+		default:
+			return ucwords( str_replace( '_', ' ', $id ) );
+	}
+}
+
+function render_page( MetadataDocument $metadata, string $tab, string $section ) {
+	iframe_header( __( 'Plugin Installation' ) );
+	render( $metadata, $tab, $section );
+	wp_print_request_filesystem_credentials_modal();
+	wp_print_admin_notice_templates();
+
+	iframe_footer();
+}
+
+/**
+ * Displays plugin information in dialog box form.
+ *
+ * @since 2.7.0
+ *
+ * @global string $tab
+ */
+function render( MetadataDocument $doc, string $tab, string $section ) {
+	$sections = (array) $doc->sections;
+
+	if ( ! isset( $sections[ $section ] ) ) {
+		$section = array_keys( $sections )[0];
+	}
+
+	$releases = $doc->releases;
+	usort( $releases, fn ( $a, $b ) => version_compare( $b->version, $a->version ) );
+	$latest = ! empty( $releases ) ? reset( $releases ) : null;
+
+	$_with_banner = '';
+
+	?>
+	<div id="plugin-information-scrollable">
+		<div id="<?= esc_attr( $tab ) ?>-title" class="<?= esc_attr( $_with_banner ) ?>">
+			<div class='vignette'></div>
+			<h2><?= esc_html( $doc->name ) ?></h2>
+		</div>
+		<div id="<?= esc_attr( $tab ) ?>-tabs" class="<?= esc_attr( $_with_banner ) ?>">
+			<?php
+			foreach ( $sections as $section_id => $content ):
+				$class = ( $section_id === $section ) ? ' class="current"' : '';
+				$href = add_query_arg( [
+					'tab'     => $tab,
+					'section' => $section_id,
+				] );
+				?>
+				<a
+					name="<?= esc_attr( $section_id ) ?>"
+					href="<?= esc_url( $href ) ?>"
+					<?= $class ?>
+				><?= esc_html( get_section_title( $section_id ) ) ?></a>
+			<?php
+			endforeach;
+			?>
+		</div>
+		<div id="<?= $tab; ?>-content" class="<?= esc_attr( $_with_banner ); ?>">
+			<?php render_fyi( $doc, $latest ) ?>
+
+			<div id="section-holder">
+			<?php
+			// check_requirements( $latest );
+			foreach ( $sections as $section_id => $content ) {
+				$prepared = sanitize_html( $content );
+				$prepared = links_add_target( $prepared, '_blank' );
+
+				printf(
+					'<div id="section-%s" class="section" style="display: %s;">%s</div>',
+					esc_attr( $section_id ),
+					( $section_id === $section ) ? 'block' : 'none',
+					$prepared
+				);
+			}
+			?>
+			</div>
+		</div>
+	</div>
+
+	<div id="<?= esc_attr( $tab ) ?>-footer">
+		<?php
+		if ( current_user_can( 'install_plugins' ) || current_user_can( 'update_plugins' ) ) {
+			$data = [];
+			$button = wp_get_plugin_action_button( $doc->name, $data, true, true );
+			$button = str_replace( 'class="', 'class="right ', $button );
+
+			if ( ! str_contains( $button, _x( 'Activate', 'plugin' ) ) ) {
+				$button = str_replace( 'class="', 'id="plugin_install_from_iframe" class="', $button );
+			}
+
+			echo wp_kses_post( $button );
+		}
+		?>
+	</div>
+	<?php
+}
+
+function name_requirement( string $requirement ) : string {
+	switch ( true ) {
+		case ( $requirement === 'env:wp' ):
+			return __( 'WordPress', 'fair' );
+
+		case ( $requirement === 'env:php' ):
+			return __( 'PHP', 'fair' );
+
+		case str_starts_with( $requirement, 'env:php-' ):
+			return substr( $requirement, 8 );
+
+		default:
+			return $requirement;
+	}
+}
+
+function render_fyi( MetadataDocument $doc, ReleaseDocument $release ) : void {
+	var_dump( $release );
+	?>
+	<div class="fyi">
+		<ul>
+			<?php if ( ! empty( $release ) ) : ?>
+				<li><strong><?= __( 'Version:' ); ?></strong> <?= $release->version; ?></li>
+			<?php endif; ?>
+			<?php if ( ! empty( $doc->slug ) ) : ?>
+				<li><strong><?= __( 'Slug:' ); ?></strong> <?= $doc->slug; ?></li>
+			<?php endif; ?>
+			<?php if ( ! empty( $release->requires ) ) : ?>
+				<li>
+					<strong><?= __( 'Requires:' ); ?></strong>
+					<ul>
+						<?php foreach ( (array) $release->requires as $type => $constraint ) : ?>
+							<li><?= esc_html( name_requirement( $type ) ); ?> <?= esc_html( $constraint ); ?></li>
+						<?php endforeach; ?>
+					</ul>
+				</li>
+			<?php endif; ?>
+			<?php if ( ! empty( $release->suggests ) ) : ?>
+				<li>
+					<strong><?= __( 'Suggests:' ); ?></strong>
+					<ul>
+						<?php foreach ( (array) $release->suggests as $type => $constraint ) : ?>
+							<li><?= esc_html( name_requirement( $type ) ); ?> <?= esc_html( $constraint ); ?></li>
+						<?php endforeach; ?>
+					</ul>
+				</li>
+			<?php endif; ?>
+		</ul>
+		<?php
+		if ( ! empty( $doc->authors ) ) :
+			?>
+			<h3><?= __( 'Authors' ); ?></h3>
+			<ul class="contributors">
+				<?php
+				foreach ( (array) $doc->authors as $author ) {
+					if ( empty( $author->name ) ) {
+						continue;
+					}
+					$url = $author->url ?? ( $author->email ? 'mailto:' . $author->email : null );
+					printf(
+						'<a href="%s" target="_blank">%s</a>',
+						esc_url( $url ),
+						esc_html( $author->name )
+					);
+				}
+				?>
+			</ul>
+		<?php endif ?>
+	</div>
+	<?php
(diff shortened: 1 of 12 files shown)


afragen:
While it says installing latest version - it's not. ☹️ 

But we can fix.

rmccue:
I actually did already fix that in 170d228ec311ebcc302bbd28babf1f84957905da - the video is outdated :)

afragen:
If no version specified should we default to most current version and not just wp_die()?

rmccue:
There's actually no way to get to the page without specifying a version, since the URL generation function requires a `ReleaseDocument`.

afragen:
Updated for a few issues I found.

afragen:
Need to deactivate Install button if plugin installed.

afragen:
I think we will need a consensus on how to manage slugs for installation and updating. The changes I made use the dot org standard of the slug, but we need to think longer term about possible duplicates.

afragen:
Last commit changes install slug to `slug-did` without the `did:plc|web` parts.

afragen:
A couple of PRs for here, #114 and #115

costdev:
PHPCS failures are expected. @afragen has been working to clean those up. I'm just merging `main` ahead of his fixes coming in.

inc/packages/admin/namespace.php
```diff
@@ -0,0 +1,189 @@
...
+		wp_die( __( 'No version specified for the plugin.', 'fair' ) );
+	}
+
+	$skin = new \WP_Upgrader_Skin();
+	$res = Packages\install_plugin( $id, $version, $skin );
+	var_dump( $res );
+	exit;
+}
+
+function embedded_info_page() {
```
afragen: Is this function used anywhere?

afragen: Does this function require nonce checking?

inc/packages/admin/namespace.php
```diff
@@ -0,0 +1,189 @@
...
+
+	// If the plugin is not a FAIR package, do nothing.
+	if ( ! preg_match( '/^did:(web|plc):.+$/', $_REQUEST['plugin'] ) ) {
+		return;
+	}
+
+	maybe_hijack_plugin_info();
+}
+
+function maybe_hijack_plugin_info() {
```
afragen, line 250: Does this function require nonce checking?

inc/packages/admin/info.php
```diff
@@ -205,6 +236,13 @@ function render_banner( ReleaseDocument $release ) {
 	<?php
 }
 
+/**
+ * Name requirement
```
costdev:
```suggestion
 * Name requirement.
```

inc/packages/namespace.php
```diff
@@ -0,0 +1,291 @@
...
+
+	switch ( $parts[1] ) {
+		case PLC::TYPE:
+			return new PLC( $id );
+
+		case Web::TYPE:
+			return new Web( $id );
+
+		default:
+			return new WP_Error( 'fair.packages.validate_id.invalid_type', __( 'Unsupported DID type.', 'fair' ) );
```
```suggestion
			return new WP_Error( 'fair.packages.validate_did.invalid_method', __( 'Unsupported DID method.', 'fair' ) );
```
1. Consistently using `validate_did`.
2. Changing from `type` to `method` to match the spec's DID syntax.
  - For the same reason, should change `TYPE` to `METHOD` in the `PLC` and `Web` classes as we discussed.

inc/packages/did/class-web.php
```diff
@@ -0,0 +1,56 @@
...
+ */
+class Web implements DID {
+	const TYPE = 'web';
+
+	/**
+	 * Decentralized ID.
+	 *
+	 * @var string
+	 */
+	protected string $id;
```
costdev, line 21: I think we should rename this to `$did` as well as other `id` references (`$id`, `get_id()`) in the `DID` interface and its implementing classes (PLC and Web at the moment).

The DID as the `id` pertains to some (not all) documents/properties per the spec, but a DID itself isn't its own ID; that's the method-specific-id. Since both DID and method-specific-id contain "id", rather than use `$id` at all in the DID interface and its implementing classes (PLC and Web), using `$did` makes it clear.

rmccue: I'm -0 on renaming this really (as mentioned); the DID _is_ the ID, and the MSID is some arbitrary thing with meaning only to the method. If you feel strongly, we can rename it, but I'd prefer not personally.

costdev: It's not so much that I feel strongly about it -  Rather, I think the main thing for me is that since there are two things containing `id` when it comes to a DID - the DID itself and the MSID component - it just seems more intuitive and simpler to me that we'd call the property `did` so it's immediately clear what it is. So that I have the full picture, what was the reason for calling the property `$id` rather than `$did` for the interface and implementing classes?

afragen: I don't think we're using the MSID component for anything.

costdev: Oh I know we're not storing the method-specific-id separately and we may never do so.

I meant that anyone coming in who knows about DID syntax will see a DID interface and implementing classes with properties for a method and an ID, and `id` is ambiguous - at first reading, it could be the DID or the method-specific-id to be appended after the method. It requires reading the docblock to confirm what `id` actually is, and my initial thought is that I don't think that should be necessary.

But since I don't know the reason why `id` was chosen over `did` for the property name, I'm only seeing part of the picture, so my question ultimately boils down to "`id` instead of `did` - how come?"

inc/packages/namespace.php
```diff
@@ -0,0 +1,310 @@
...
+
+/**
+ * Install a plugin from a FAIR DID.
+ *
+ * @param string $id DID of the package to install.
+ * @param string|null $version Version to install. If null, the latest version is installed.
+ * @param WP_Upgrader_Skin $skin Plugin Installer Skin.
+ * @return bool|WP_Error True on success, WP_Error on failure.
+ */
+function install_plugin( string $id, ?string $version = null, $skin ) {
```
afragen: I'm seeing this on PHP 8.3

```
PHP Deprecated:  Optional parameter $version declared before required parameter $skin is implicitly treated as a required parameter in /Users/afragen/Documents/github/fair-plugin/inc/packages/namespace.php on line 133
```
 FWIW this is locally.

costdev: So we'll need to swap version and skin, or default skin to null and handle that inside the function.

The latter is preferable if possible.

afragen: Or just ensure that when we call `install_plugin()` we ensure a `$version` parameter.

costdev: True, I just assumed that version being optional was important, but if it's not then that's fair enough.

afragen (approved):

afragen:
Not yet fully complete, but nothing stopping from merging.

- update needed for Install modal
- need to get/install `did:web` data

afragen (approved):

inc/packages/did/class-did.php
```diff
@@ -0,0 +1,32 @@
...
+
+namespace FAIR\Packages\DID;
+
+interface DID {
+	/**
+	 * Get the DID type.
+	 *
+	 * One of plc, web.
+	 */
+	public function get_type() : string;
```
costdev: @rmccue Are we changing this to `get_method()` and the corresponding constant in PLC/Web to `METHOD` for consistency with the spec? Or?

inc/packages/admin/namespace.php
```diff
@@ -0,0 +1,257 @@
...
+			<input
+				type="text"
+				id="plugin_id"
+				name="plugin_id"
+				pattern="did:(web|plc):.+"
+				placeholder="did:..."
+			/>
+			<?php submit_button( _x( 'View Details', 'plugin', 'fair' ), '', '', false ); ?>
+		</form>
+		<p class="fair-direct-install__note">
```
joedolson: This text is important to knowing how to fill out the plugin ID field; let's give it an explicit association with the input field via `aria-describedby`.

inc/packages/class-upgrader.php
```diff
@@ -0,0 +1,691 @@
...
+	 * @param ReleaseDocument $release Release document.
+	 * @param  bool $clear_cache Whether to clear the plugin updates cache if successful.
+	 *                                    Default true.
+	 * @param  bool $overwrite Whether to overwrite plugin folder.
+	 *                                    Default true.
+	 * @return bool|WP_Error True if the installation was successful, false or a WP_Error otherwise.
+	 */
+	public function install( MetadataDocument $package, ReleaseDocument $release, $clear_cache = true, $overwrite = false ) {
+		$this->init();
+		// $this->install_strings();
```
costdev: Any reason why this is commented out? This is what makes the skin use the actual strings instead of just the codes.

Edit: Just in case, apologies if the tone of "Any reason" sounded hostile. I meant it casually.

inc/packages/class-upgrader.php
```diff
@@ -0,0 +1,691 @@
...
+			case 'wp-plugin':
+				return $this->install_plugin( $clear_cache, $overwrite );
+
+			case 'wp-theme':
+				return $this->install_theme( $clear_cache, $overwrite );
+
+			default:
+				return new WP_Error( 'fair.packages.upgrader.install.invalid_type', 'Invalid package type.' );
+		}
+	}
```
afragen, line 460:
```suggestion
	}

```

inc/packages/class-upgrader.php
```diff
@@ -0,0 +1,691 @@
...
+	/**
```
afragen:
```suggestion
	}

	/**
```

inc/packages/admin/namespace.php
costdev: Done in [83e092f](https://github.com/fairpm/fair-plugin/pull/71/commits/83e092ff8e176ae585764153fcf4f27aa7007959).

costdev:
Updated video showing AJAX installing and using two plugins, one that uses a GitHub release asset and one that doesn't, showing the install process looks exactly the same.

https://github.com/user-attachments/assets/76774239-271f-454e-b094-4e4dd9b6a107

inc/packages/admin/namespace.php
```diff
@@ -0,0 +1,312 @@
...
+			>Plugin ID</label>
+			<input
+				type="text"
+				id="plugin_id"
+				name="plugin_id"
+				pattern="did:(web|plc):.+"
+				placeholder="did:..."
+				aria-describedby="fair-direct-install__note"
+			/>
+			<?php submit_button( _x( 'View Details', 'plugin', 'fair' ), '', '', false ); ?>
```
joedolson, line 137: This button is active before a plugin ID has been entered. It should be disabled when there is no DID or the DID is known to be invalid.

inc/packages/admin/namespace.php
```diff
@@ -0,0 +1,312 @@
...
+			<?= __( 'Plugin IDs should be in the format <code>did:web:...</code> or <code>did:plc:...</code>', 'fair' ); ?>
+		</p>
+	</div>
+	<script>
+		// On submit, trigger the thickbox with the plugin information.
+		document.querySelector( '.fair-direct-install__form' ).addEventListener( 'submit', function ( e ) {
+			e.preventDefault();
+			e.stopPropagation();
+			const id = document.getElementById( 'plugin_id' ).value;
+
```
joedolson, line 149: I'd get this constant outside of the event listener so it can be used as a conditional to prevent execution or trigger an error message when there's no DID added.

inc/packages/admin/namespace.php
joedolson: Or it should be enabled, but trigger an error if no DID is added. Either is fine by me.

joedolson: Simplest solution might just be to mark the input as required.

inc/packages/admin/namespace.php
```diff
@@ -0,0 +1,312 @@
...
+/**
+ * Render direct installer tab.
+ *
+ * @return void
+ */
+function render_tab_direct() {
+	?>
+	<div class="fair-direct-install">
+		<p class="fair-direct-install__help">
+			<?= __( "If you have a plugin's ID, enter it here to view the details<br />and install it.", 'fair' ); ?>
```
joedolson: This text seems unnecessarily verbose. Can it just be "Enter a plugin's DID to view it's details and install."

Also, let's omit the manual `br`. That will create an awkward pause for screen reader users.

inc/packages/did/class-plc.php
```diff
@@ -0,0 +1,79 @@
...
+
+		if ( is_wp_error( $response ) ) {
+			return $response;
+		}
+
+		$data = json_decode( $response['body'] );
+		if ( json_last_error() !== JSON_ERROR_NONE ) {
+			return new WP_Error( 'fair.packages.did.json_error', __( 'Unable to parse DID document response.', 'fair' ) );
+		}
+
```
joedolson, line 76: If you pass an invalid DID, the `$response` will be valid, but not contain a document. Need to handle the error message returned here.

joedolson (changes requested):
A couple UX issues, and an unhandled error path with invalid DIDs.

inc/packages/admin/namespace.php
joedolson: Making the input field required makes this largely irrelevant. If we later add some kind of pre-submission validation, it could be useful, but at the moment I don't think it's relevant.

inc/packages/admin/namespace.php
rmccue: No; the idea originally was for this to display the info page, but I discovered the behaviour was tied heavily into the iframe/modal, so skipped it.

inc/packages/did/class-web.php
rmccue: I used `id` since I tend to use `id` universally as the property for something that is the primary key/unique identifier for a model. The DID string is the unique identifier for the `DID` object, just as post IDs are the unique identifier for posts.

> at first reading, it could be the DID or the method-specific-id to be appended after the method

I don't see this as confusing personally, but ymmv. The MSID is not the ID, in the same way that `/foo/bar` is not a full URI without the protocol and host attached. Calling the MSID the ID would be way more confusing, and as per other discussions I don't think we ever really want to be dealing with the separated MSID.

inc/packages/did/class-did.php
rmccue: Changing to `method` consistently makes sense to me; I picked `type` without really thinking about it, but sticking to the standard term is better.

inc/packages/admin/info.php
```diff
@@ -0,0 +1,474 @@
...
+ */
+
+namespace FAIR\Packages\Admin\Info;
+
+use FAIR\Packages;
+use FAIR\Packages\Admin;
+use FAIR\Packages\MetadataDocument;
+use FAIR\Packages\ReleaseDocument;
+
+use function FAIR\Updater\get_packages;
```
rmccue: This should be `use` on the namespace rather than the function.

inc/updater/namespace.php
```diff
@@ -1,30 +1,88 @@
...
+ *
+ * @param array  $args Array of http args.
+ * @param string $url  Download URL.
+ *
+ * @return array
+ */
+function maybe_add_accept_header( $args, $url ) : array {
+	$releases = wp_cache_get( RELEASE_PACKAGES_CACHE_KEY ) ?: [];
+
+	if ( ! str_contains( $url, 'api.github.com' ) ) {
```
rmccue: I really don't want to add this exception for a specific source if we can at all avoid it, and I'm pretty certain we can with a properly constructed standard Accept header.

`Accept` indicates data types in preferential order (as indicated by `q=`), so we should be able to use a "universal" Accept header, similar to browsers. Browsers send `text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8` eg

For example, `Accept: application/zip;q=1.0, application/octet-stream;q=0.7, */*;q=0.1` would indicate we accept zips, then binary streams, then anything (in that order).

I'll test this out.

inc/packages/admin/namespace.php
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	$did = 'did:' . explode( '-did:', str_replace( '--', ':', $slug ), 2 )[1];
+	if ( ! preg_match( '/^did:(web|plc):.+$/', $did ) ) {
+		return $result;
+	}
+
+	wp_cache_set( ACTION_INSTALL_DID, $did );
+	Updater\add_package_to_release_cache( $did );
+	add_filter( 'http_request_args', 'FAIR\\Updater\\maybe_add_accept_header', 20, 2 );
```
rmccue: Given these functions aren't specific to the updater, and we're reusing them here, we should move them upwards into the Packages namespace imo.

inc/packages/admin/namespace.php
afragen: So we can delete?

inc/updater/namespace.php
afragen: Fails in testing.

inc/packages/did/class-did.php
```diff
@@ -0,0 +1,32 @@
...
+ * DID Interface.
+ *
+ * @package FAIR
+ */
+
+namespace FAIR\Packages\DID;
+
+interface DID {
+	/**
+	 * Get the DID type.
```
afragen:
```suggestion
	 * Get the DID method.
```

inc/updater/namespace.php
rmccue: This is caused on the _other_ side of things, where `api.github.com` is incorrectly wanting `Accept: application/json` despite giving a zip response. I've contacted GitHub Support.

It looks like GitHub are also not parsing the `q` parameter correctly, so `*/*;q=0.1` or `*/*;q=0` incorrectly receive priority over `application/octet-stream`. This is noncompliant with the HTTP spec afaik.

PR #84, opened 2025-06-12
Add fair-de_DE.l10n.php

Presskopp:
.. because I'm helping @La-Geek a bit, who did the translation, but asked me to create the according file.


La-Geek:
Thank you @Presskopp :)

PR #91, opened 2025-06-16
Estonian translation for FAIR plugin

webdocdisain:
Latest translation files to estonian language


timiwahalahti:
Thanks @webdocdisain!

PR #92, opened 2025-06-16
Add Dutch translations files (including formal)

toineenzo:
Based on 0.2.0
Includes both normal and formal translations with PO, MO, l10n.php and json files

Signed-off-by: Toine Rademacher (toineenzo) <hi@toine.zip>


timiwahalahti:
Thanks @toineenzo!

PR #96, opened 2025-06-16
Fix loading translations

TeemuSuoranta:
Loading localization files was not happening as `load_plugin_textdomain` was not called while translations are stored under the plugin.

Without this addition none of the translations are loaded.

Tested with fi translation.


Ipstenu (approved):
Approving even though code freeze because it's a break/fix

PR #97, opened 2025-06-16
Make priority later to override GU data, pref  to GUL if present

afragen:
Needed if GU running in same site.


afragen:
I have tested and at least in one case this was necessary as GU's data overrode GUL's data. In GUL the update is listed as a release asset and from the header info it isn't.

costdev (approved):

PR #98, opened 2025-06-16
Add generate missing i18n files action

timiwahalahti:
On today's Weekly Tech Call, attendees discussed about translation files. We agreed that we should probably generate MO, PHP, and JSON files to allow contributors to submit translation PRs only with PO files, as well as to ensure that we always have PHP and JSON versions.

This PR generates those files every time the PO's get updated in the main branch.


afragen:
Let’s merge and make sure it works as expected. LGTM

PR #100, opened 2025-06-17
Update fi.po timestamp to activate generate-missing-i18n-files workflow

timiwahalahti:
Updating the PO file timestamp should be enough to activate recently added generate-missing-i18n-files workflow.


timiwahalahti:
Was able to run the action by manual trigger, which I thought would be impossible until it has run the first time.

PR #103, opened 2025-06-17
Add `.gitattributes` and exclude other development files.

### LINKED ISSUE #102: Excluded files aren't excluded from the plugin-only ZIP release
When the plugin is bundled with Core, the contents of .distignore are excluded from it.

However, when the plugin-only ZIP is generated in GitHub Actions using `git archive -o`, it has no such exclusions.

To add these exclusions, we can use a `.gitattributes` file with the same contents as `.distignore`, with each line ending with `export-ignore`.

costdev:
This adds a `.gitattributes` file so that packages created using `git archive -o` will correctly exclude files that should not be included in the release package. `.distignore` has been updated to include `.gitattributes`. Additional development files, such as `.wp-env.json` and `.editorconfig` have also been added to both ignore lists.

Fixes #102

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -8,6 +8,8 @@ phpunit.xml.dist
 wp-tests-config-sample.php
 composer.json
 composer.lock
+package.json
+package-lock.json
 .wp-env.json
 .gitignore
 .gitattributes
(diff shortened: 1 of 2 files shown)


afragen (approved):

costdev:
Before:
image

After:
image

PR #105, opened 2025-06-17
Use "fair-plugin" in the bundled distribution script.

### LINKED ISSUE #104: The directory name differs between bundled and plugin-only ZIPs
When the plugin is bundled with Core, it's in a `fair` directory, whereas the plugin-only ZIP generated for a release is in a `fair-plugin` directory. As these serve as the plugin's slug, the directory names should be consistent.

I'm proposing that we rename the bundled plugin directory to `fair-plugin` to match the repository's name.

costdev:
This replaces instances of `fair` with `fair-plugin` in the `bin/bundle.sh` script for consistent naming of the package and directory.

Fixes #104

### DIFF
--- a/.github/workflows/releases.yml
+++ b/.github/workflows/releases.yml
@@ -22,7 +22,7 @@ jobs:
         run: echo "tag=${GITHUB_REF#refs/tags/}" >> $GITHUB_OUTPUT
 
       - name: Build project
-        run: git archive -o /tmp/${{ github.event.repository.name }}-${{ steps.tag.outputs.tag }}.zip --prefix=${{ github.event.repository.name }}/ ${{ steps.tag.outputs.tag }}
+        run: git archive -o /tmp/fair-${{ steps.tag.outputs.tag }}.zip --prefix=fair/ ${{ steps.tag.outputs.tag }}
 
       - name: Create WP distribution files
         run: bin/bundle.sh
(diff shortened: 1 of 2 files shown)


afragen:
`fair-plugin` is derived from the repository name as this is the slug used when installing the plugin in `wp-content/plugins/`

Is the goal or result to install the plugin as `wp-content/plugins/fair`?

costdev:
So we have two possible approaches (without adding more):
1. Rename the `fair-plugin-X.X.X.zip` and its internal `fair-plugin` directory to `fair`
    - Both renamed for consistency, but the internal directory name is the actual issue.
2. Rename the `fair` directory in the distributions to `fair-plugin`.

I opted for the first simply because `-plugin` in a plugin directory name is redundant. Also note that the distributions script doesn't know the repository name to enforce consistency.

I don't have a strong opinion on it though.

afragen:
If the goal is to change `fair-plugin` to `fair`, the simplest way is to rename the repository. Otherwise there will be all sorts of potential issues when scraping the GitHub API for data.

costdev:
Since we have `fairpm/fair-protocol` etc, `fairpm/fair` isn't ideal, so instead of renaming the repository, I went for option 2: Renaming the `fair` directory in the bundled ZIPs to `fair-plugin` to match the repository name and the plugin-only ZIP's internal directory name.

costdev:
Before:
image

After:
image

PR #109, opened 2025-06-17
Update sv_SE to .pot for 2025-06-16

tobifjellner:


timiwahalahti:
Thanks!

I'm keeping this open, as there are likely going to be changes in strings for 0.4.0, so updating translations closer to the next release is probably needed.

afragen:
So we only really need the .po file as the rest are created via a GitHub Action here.

It also seems that there are conflicts with the .po in the PR. I know it's a big ask, but we can merge as soon as this is fixed.

afragen (approved):

PR #113, opened 2025-06-18
Update ca POT translation (6-18-25)

marcarmengou:


timiwahalahti:
Thanks @marcarmengou!

I'm keeping this open, as there are likely going to be changes in strings for 0.4.0, so updating translations closer to the next release is probably needed.

fair-ca.po
```diff
@@ -0,0 +1,206 @@
...
+msgstr "Trieu la imatge del perfil"
+
+#: inc/dashboard-widgets/namespace.php:101
+msgctxt "default event location"
+msgid "Online"
+msgstr "En línia"
+
+#: inc/dashboard-widgets/namespace.php:186
+msgid "Events (by The WP World)"
+msgstr "Esdeveniments (pel WP World)"
```
joedolson: Technically, the name of the organization here is "The WP World". I'm not sure what the standard is for translating articles when they're part of the organization name.

fair-ca.po
```diff
@@ -0,0 +1,206 @@
...
+msgid "Avatars will be loaded from the selected source."
+msgstr "Es pujaran els avatars des de la font seleccionada."
+
+#: inc/version-check/namespace.php:101
+msgctxt "browser version check"
+msgid "your browser"
+msgstr "el vostre navegador"
+
+#: inc/version-check/namespace.php:100
+msgctxt "browser version check"
```
joedolson: This isn't an issue with the translation, but might be an issue in the plugin. I'm not sure whether this context message is correct. The context in which the check occurs is during the browser version check, but the string is about checking the platform. Do you think this is correct?

fair-ca.po
marcarmengou: Good catch, I will correct it.

fair-ca.po
marcarmengou: I think that, if what BrowseHappy API responds to is Windows, macOS or Linux, a context message like "operating system check" would be better.

That way it would be more explicit about what is being checked (the operating system). It would avoid confusion with the browser, which is the current problem with 'browser version check'. And it would help translators to choose accurate terms, such as "el vostre sistema operatiu" in Catalan.

PR #114, opened 2025-06-18
Installer fixes 1

afragen:
@rmccue just a couple of small fixes I found. From PHP Warnings/Errors on PHP 8.3

### DIFF
--- a/inc/packages/class-upgrader.php
+++ b/inc/packages/class-upgrader.php
@@ -428,7 +428,7 @@ class Upgrader extends WP_Upgrader {
 		}
 
 		// Get environmental requirements if set.
-		foreach ( $this->release->require as $pkg => $ver ) {
+		foreach ( $this->release->requires as $pkg => $ver ) {
 			switch ( true ) {
 				// WordPress requirements.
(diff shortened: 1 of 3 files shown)


inc/packages/class-upgrader.php
```diff
@@ -469,7 +469,7 @@ public function check_requirements( $source ) {
 					$php_ext = substr( $pkg, 8 );
 					if ( extension_loaded( $php_ext ) ) {
 						// Extension is loaded, skip.
-						continue;
+						break;
```
rmccue: Oh this was kinda intentional, shouldn't be an error?

rmccue (approved):

inc/packages/class-upgrader.php
afragen: It was strange as VSCode was shouting things like should this be `continue 2`.

I can take this out if you want and see if the error/warning recurs and screenshot it.

afragen: reverted

rmccue: Ah, then it probably should indeed be `continue 2`, but better that than `break`

PR #115, opened 2025-06-18
Need to rename plugin slug on installation

afragen:
Need to rename plugin slug on installation otherwise GitHub will name it `afragen-git-updater-<hash>`.

This seems like a good place for discussion of what to use for unique slugs. This PR uses a combination of the slug and the DID. This seems to be human readable yet unique and identifies the DID.

screenshot_663


inc/packages/class-upgrader.php
```diff
@@ -573,4 +575,28 @@ protected function validate_theme( $dir ) {
 
 		$this->new_theme_data = $info;
 	}
+
+	/**
+	 * Correctly rename dependency from git host download name.
```
costdev: "dependency from git host download name" isn't quite clear to me.

I think "dependency" is a typo for "directory", right?

While the code itself covers the git host issue, it might be worth describing it as something like this:

```php
/**
 * Renames a package's directory when it doesn't match the slug.
 *
 * This is commonly required for packages from Git hosts.
 *
 * ...
 */
```

inc/packages/class-upgrader.php
```diff
@@ -573,4 +575,28 @@ protected function validate_theme( $dir ) {
...
+	 *
+	 * @param string $source        Path of $source.
+	 * @param string $remote_source Path of $remote_source.
+	 *
+	 * @return string
+	 */
+	public function rename_source_selection( string $source, string $remote_source ) {
+			global $wp_filesystem;
+
+			$new_source = trailingslashit( $remote_source ) . $this->package->slug . '-' . explode( ':', $this->package->id )[2];
```
costdev: Not a big deal: Could be bumped below the early return (since it's not used until later) so it's closer to its usage.

inc/packages/class-upgrader.php
afragen: Yeah

costdev:
I raised some concerns I had about the deduplication naming elsewhere, so I'm just noting my thoughts on this since those were informal chats. Putting them here as this is the first deduplication implementation I've seen in the repo, so I'm not dissing the work on this PR :sweat_smile:

Functions such as `is_plugin_active()` and `is_plugin_inactive()`, and others will be affected by deduplication. This will require plugin developers to make changes to such checks. Note that these calls may be made by the author of the plugin being checked, or by any other developer including theme developers, so the awareness won't always be there.

It was mentioned that FAIR could include its own function(s). This would require adoption by every developer and tutorial writer that reference the native WP functions. Aside from this taking a large amount of time to accomplish and it not being possible to inform everyone/see them act on it, I'm not sure new functions in FAIR can adequately augment the native WP functions in terms of achieving the desired behaviour.

Before we implement _any_ deduplication, I think we need to have a detailed discussion covering:
1. An overview of what we're trying to accomplish (i.e. deduplication)
2. What is viable to ask from others to accommodate this (developers, tutorial writers, etc) - if anything at all.
3. What we could implement to achieve the goal without causing disruption or false results from checking plugin/theme status, etc.

In the meantime, I'd suggest this PR renames the directory to just the `slug` rather than `slug-did` since that should still achieve what the PR sets out to do.

afragen:
The PR can easily be modified for just the slug if that’s the consensus opinion. 

Waiting for more opinions and thoughts.

afragen:
PR changed back to using normal slug.

inc/packages/class-upgrader.php
```diff
@@ -655,4 +657,30 @@ protected function validate_theme( $dir ) {
...
+		}
+
+			$new_source = trailingslashit( $remote_source ) . $this->package->slug;
+
+		if ( trailingslashit( strtolower( $source ) ) !== trailingslashit( strtolower( $new_source ) ) ) {
+			$wp_filesystem->move( $source, $new_source, true );
+		}
+
+			return trailingslashit( $new_source );
+	}
```
costdev:
```suggestion
	public function rename_source_selection( string $source, string $remote_source ) {
		global $wp_filesystem;

		if ( basename( $source ) === $this->package->slug ) {
			return $source;
		}

		$new_source = trailingslashit( $remote_source ) . $this->package->slug;

		if ( trailingslashit( strtolower( $source ) ) !== trailingslashit( strtolower( $new_source ) ) ) {
			$wp_filesystem->move( $source, $new_source, true );
		}

		return trailingslashit( $new_source );
	}
```
Some extra indents were there.

costdev (changes requested):

costdev (approved):

PR #121, opened 2025-06-20
un-escape class-lite.php for Ryan

afragen:
Removed most all the escaping and added `use` statements.

I'm trying to make @rmccue happier. 😄


rmccue:
Given this is a vendor class we're incorporating it, does it make sense to change it at all?

afragen:
I made similar changes in afragen/git-updater-lite 😉

PR #122, opened 2025-06-23
Keep link text in translation strings

alexclassroom:


inc/settings/namespace.php
```diff
@@ -71,14 +71,13 @@ function render_settings_page() {
 		<h1><?php esc_html_e( 'FAIR Settings', 'fair' ); ?></h1>
 		<p><?php printf(
 			/* translators: %s: AspirePress website URL */
-			esc_html__( 'Your site is now disconnected from WordPress.org managed servers and uses a dedicated mirror from %s. In the coming weeks and months, we\'ll be making the mirror configurable and add in the ability to add your own.', 'fair' ),
-			'<a href="https://aspirepress.org/">AspirePress</a>'
+			esc_html__( 'Your site is now disconnected from WordPress.org managed servers and uses a dedicated mirror from %1$sAspirePress%2$s. In the coming weeks and months, we\'ll be making the mirror configurable and add in the ability to add your own.', 'fair' ),
+			'<a href="https://aspirepress.org/" target="_blank">', '</a>'
```
costdev: I don't think including `target="_blank"` should be mixed in with this string change. Opening links in a new tab is somewhat of a hot topic, so I think it's best discussed/handled elsewhere.

joedolson: Agreed. I also don't think that this specific text should be translatable; it's a proper noun, which shouldn't generally be translated.

inc/settings/namespace.php
```diff
@@ -71,14 +71,13 @@ function render_settings_page() {
...
 		); ?></p>
 		<p><?php printf(
 			/* translators: %1$s: FAIR GitHub URL, %2$s: FAQ URL */
-			esc_html__( 'If you want to learn more, check out the %1$s and our %2$s.', 'fair' ),
-			'<a href="https://fair.pm/">FAIR GitHub</a>',
-			'<a href="https://github.com/fairpm/tsc/blob/main/faqs/README.md">FAQ</a>'
+			esc_html__( 'If you want to learn more, check out the %1$sFAIR GitHub%2$s and our %3$sFAQ%2$s.', 'fair' ),
+			'<a href="https://fair.pm/" target="_blank">', '</a>', '<a href="https://github.com/fairpm/tsc/blob/main/faqs/README.md" target="_blank">'
```
joedolson: FAQ should be translatable, but not FAIR Github.

Issue #70, opened 2025-06-10
Avatar source setting should be network-wide


johnbillion:
The "Avatar Source" setting should be network-wide as it makes little sense for different sites within a Multisite installation to use different avatar providers. This also means on a network of 100 sites you don't need to change the setting 100 times.

The `fair_settings` option should be switched to a network option (which behaves as a regular option on single-site installations) and the Settings screen moved to the network admin when Multisite is enabled. If further FAIR settings are added in the future which should remain as per-site settings then they can go into a separate regular option.

norcross:
i can tackle this.

rmccue:
Ideally, this setting should be on the Discussion screen imo along with other avatar settings.

norcross:
ok @rmccue so the discussion settings don't exist at the network level. all we have there is "Network Settings". so for single sites, we can get it into the discussion settings no problem. for multisite, i'll have to get it into there

rmccue:
Makes sense, I think just avoiding having a full screen either way for the like one or two settings we have.

joedolson:
This should be closed, I believe. @norcross Can you confirm that?

norcross:
@joedolson yessir!

Ipstenu:
I’m going to close this because we can’t put it on network pages without making  new page. That said, we should talk about that too!

PR #123, opened 2025-06-23
refactoring avatar settings

norcross:
this PR updates the current settings implementation for the avatar source. this breaks up the settings and moves to two separate functions.

* `load_single_site_avatar_settings` is for single sites, and fully utilizes the Settings API
* `wpmu_options` and `update_wpmu_options` to handle the multisite

the setting key is currently set as a serialized array, and this converts it to a single key. this also removes the CSS file that was added for the stand-alone settings page.

this addresses #70


afragen:
When this PR used with a single site, the FAIR Settings link is not displayed.

inc/avatars/namespace.php
```diff
@@ -2,12 +2,13 @@
...
 
+const AVATAR_SRC_SETTING_KEY = 'fair_avatar_source';
+
 /**
  * Bootstrap.
  */
 function bootstrap() {
-	$options = get_option( 'fair_settings', [] );
-	$avatar_source = array_key_exists( 'avatar_source', $options ) ? $options['avatar_source'] : 'fair';
+	$avatar_source = get_site_option( AVATAR_SRC_SETTING_KEY, 'fair' );
```
joedolson, line 16: It seems like this is just going to ignore existing settings. That might be fine at this point, with low usage, but I think that single sites should migrate prior settings.

joedolson (changes requested):
We should do something to handle previous settings.

inc/settings/namespace.php
```diff
@@ -7,179 +7,133 @@
 
 namespace FAIR\Settings;
 
+const AVATAR_SRC_SETTING_KEY = 'fair_avatar_source';
```
johnbillion: Minor: Instead of duplicating this here you can `use` it from the `FAIR\Avatars` namespace.

inc/avatars/namespace.php
norcross: i had that thought myself, but is there anything in place for a "run once" update logic?

joedolson: Not yet - but if we're going to change how we do things, we're going to need that. It can just be an activation hook. 

On that note, we also need an uninstall.php, so that we can remove our settings. We overlooked that in the original implementation.

norcross: let's open a different ticket for the uninstall, but i've got the settings fix included in this PR now.

PR #126, opened 2025-06-24
Introduce a Git hook to help ensure commits are signed-off

philipjohn:
Adds a commit-msg hook that checks for the presence of the "Signed-off-by: " string in the commit message and rejects the commit when it's missing. Additonally, it checks if there is more than one "Signed-off-by: " string and rejects that too (this is from the default hook sample).

To make life easier for contributors a composer script is added so that running `composer run setup` runs a simple bash script (`bin/install-git-hooks.sh`) which adds a symlink from the committed git hook to `.git/hooks/commit-msg`.

Finally, the docs are updated to let people know they can install the git hook. I opted to leave the manual instructions for signing commits in place, as it's helpful for understanding as well.

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+# Ensure the commit is signed-off by the author.
+if ! grep -q '^Signed-off-by: ' "$1"; then
+	echo >&2 "Commit message must be signed off by the author."
+	exit 1
+fi
+
+# Catch duplicate Signed-off-by lines.
+test "" = "$(grep '^Signed-off-by: ' "$1" |
+	 sort | uniq -c | sed -e '/^[ 	]*1[ 	]/d')" || {
+	echo >&2 Duplicate Signed-off-by lines.
+	exit 1
+}
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,9 @@
         "lock": false
     },
     "scripts": {
+		"setup": [
+			"./bin/install-git-hooks.sh"
+		],
         "test": [
             "Composer\\Config::disableProcessTimeout",
             "@php ./vendor/phpunit/phpunit/phpunit"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,6 +14,8 @@ $ git commit -s -m 'My commit message.'
 
 **Please Note:** This is adding a _sign-off_ to the commit, which is not the same as *signing* your commits (which involves GPG keys).
 
+Alternatively, run `composer run setup` to install a git hook which will automatically reject your commits if they are not signed off.
+
 ## Development Environment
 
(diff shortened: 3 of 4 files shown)


rmccue:
I'm generally against bundling this into the repo (eg with things like Husky), but I think a manual command is an acceptable tradeoff. (Note though, this requirement is true for all repositories on this organisation, and indeed as a general rule across LF projects.)

Should we add [a prepare-commit-msg hook](https://stackoverflow.com/a/46536244/2575) as well? Git does intentionally not have a config flag for this as "Adding the Signed-off-by trailer to a patch should be a conscious act and means that you certify you have the rights to submit this work under the same open source license.", but I know a lot of people are using it.

Ipstenu:
I think having this as a check will help make sure people actually do the thing before we get to the part where we have to tell them "You need to rebase the PR so we can accept it." 

Better link for Ryan's link - https://stackoverflow.com/a/46536244

@philipjohn Any thoughts?

composer.json
```diff
@@ -19,6 +19,9 @@
         "lock": false
     },
     "scripts": {
+		"setup": [
+			"./bin/install-git-hooks.sh"
```
kasparsd, line 23: 1. I wonder if this would be better suited for `package.json` since that is the canonical runner of scripts.

2. There is `git config --local core.hooksPath .githooks` which would set the hooks directory and avoid the need to symlink files. This makes it work across operating systems and it is the same logic that [husky uses behind the scenes](https://github.com/typicode/husky/blob/ec10d87dbb7e2fe83115b83fc72566f0e64dff2c/index.js#L14). It makes is clear which hooks are available and can be enabled.

.githooks/commit-msg
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# Ensure the commit is signed-off by the author.
+if ! grep -q '^Signed-off-by: ' "$1"; then
```
kasparsd, line 4: Could we rewrite this in NodeJS to remove the Bash dependency? All of the tooling already requires Node so that would work out of the box for everyone consistently.

chuckadams:
I'm 👍 on a `composer setup` command, but I'm with @rmccue that we should just install the `prepare-commit-msg` hook rather than check for the signoff.  

I'd also say if we're ever planning to do more with git hooks, since we're already composer-based we should look at https://github.com/captainhook-git/captainhook.  I'd call it overkill for just this though.

kasparsd:
@cdils it appears this was closed because the target branch was deleted. Do we want to reopen this and adjust the branch?

Were there other pull requests targeting the same release_1.4.0 branch which got closed?

cdils:
@kasparsd Thank you for the catch. We thought we had stopped the release branch (`release_1.4.0` in this case) from auto-deleting when merged into `development`, but apparently not. 😬

PR #138, opened 2025-06-28
Linting updates for #71

afragen:
Coding standards updates for  #71 

Very much needs review.

### DIFF
--- a/inc/packages/admin/namespace.php
+++ b/inc/packages/admin/namespace.php
@@ -182,7 +182,7 @@ function handle_direct_install() {
 }
 
 /**
- * Hijack embedded inof page.
+ * Hijack embedded info page.
  *
  * @return void
  */


inc/packages/admin/namespace.php
```diff
@@ -151,38 +181,48 @@ function handle_direct_install() {
...
-	$id = wp_unslash( $_REQUEST['plugin'] );
+	$id = sanitize_text_field( wp_unslash( $_REQUEST['plugin'] ) );
 	if ( ! preg_match( '/^did:(web|plc):.+$/', $id ) ) {
 		return;
 	}
 
 	$metadata = Packages\fetch_package_metadata( $id );
 	if ( is_wp_error( $metadata ) ) {
-		wp_die( $metadata->get_error_message() );
+		wp_die( esc_attr( $metadata->get_error_message() ) );
```
costdev: Why `esc_attr()` rather than `esc_html()`?

inc/packages/class-upgrader.php
```diff
@@ -251,20 +304,32 @@ protected function run_install( string $destination, $options ) {
...
 		return [];
 	}
 
+	/**
+	 * Install plugin.
+	 *
+	 * @param  bool $clear_cache Boolean to clear cache.
+	 * @param  bool $overwrite Boolean to overwrite.
+	 *
+	 * @return bool|WP_Error True if the installation was successful, false or a WP_Error object otherwise.
```
costdev, line 331: `WP_Upgrader::$result` is documented as `array|WP_Error` [Ref](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-admin/includes/class-wp-upgrader.php#L93).

Is the parameter doc in Core incomplete, or should this be changed to this?
```
* @return bool|array|WP_Error True if the installation was successful, an array or a WP_Error object otherwise.
```

inc/packages/admin/info.php
```diff
@@ -58,29 +63,45 @@ function sanitize_html( string $html ) : string {
 	return wp_kses( $html, $allowed );
 }
 
+/**
+ * Get seciton title.
```
costdev:
```suggestion
 * Get section title.
```

inc/packages/class-upgrader.php
afragen: I think it's actually [here](https://github.com/WordPress/wordpress-develop/blob/99499d88a83fb7d5d0b66aa49c6b3108e6bd78df/src/wp-admin/includes/class-plugin-upgrader.php#L118).

inc/packages/admin/info.php
```diff
@@ -115,29 +139,29 @@ function render( MetadataDocument $doc, string $tab, string $section ) {
...
 				?>
 				<a
-					name="<?= esc_attr( $section_id ) ?>"
-					href="<?= esc_url( $href ) ?>"
-					<?= $class ?>
-				><?= esc_html( get_section_title( $section_id ) ) ?></a>
-			<?php
+					name="<?= esc_attr( $section_id ); ?>"
+					href="<?= esc_url( $href ); ?>"
+					<?= esc_attr( $class ); ?>
```
costdev: This will change: ` class="current'"` to ` class=current`.

image

inc/packages/admin/info.php
```diff
@@ -151,22 +175,22 @@ function render( MetadataDocument $doc, string $tab, string $section ) {
 					'<div id="section-%s" class="section" style="display: %s;">%s</div>',
 					esc_attr( $section_id ),
 					( $section_id === $section ) ? 'block' : 'none',
-					$prepared
+					esc_attr( $prepared )
```
costdev: This will sanitize the HTML content as if it was attributes.

inc/packages/admin/info.php
afragen: How about `sanitize_text_field()`?

inc/packages/did/class-plc.php
```diff
@@ -36,6 +50,11 @@ public function get_id() : string {
 		return $this->id;
 	}
 
+	/**
+	 * Fetch PLC document.
+	 *
+	 * @return stdClass|Document
```
costdev: Shouldn't this be `Document|WP_Error`?

inc/packages/admin/info.php
afragen: Should we run it back through `sanitize_html()`?

inc/packages/did/class-plc.php
afragen: you're correct

inc/packages/admin/info.php
costdev: Either that or `esc_html()` I guess.

inc/packages/admin/info.php
costdev: Same output.

costdev: Maybe just do:

```php
<?php if ( $section_id === $section ) : ?>
class="current"
<?php endif; ?>
```

Don't be surprised if an embedded PHP sniff throws an error. We could ignore that for this line/file if needs be.

inc/packages/admin/info.php
costdev: Or not `esc_html()` :joy: What was I thinking saying that? It's bloody HTML!

inc/packages/admin/info.php
```diff
@@ -151,22 +175,22 @@ function render( MetadataDocument $doc, string $tab, string $section ) {
 					'<div id="section-%s" class="section" style="display: %s;">%s</div>',
 					esc_attr( $section_id ),
 					( $section_id === $section ) ? 'block' : 'none',
-					$prepared
+					sanitize_html( $prepared )
```
costdev, line 178:
```suggestion
					// phpcs:ignore HM.Security.EscapeOutput.OutputNotEscaped -- sanitize_html() is a custom wrapper for wp_kses().
					sanitize_html( $prepared )
```

inc/packages/class-upgrader.php
costdev: Strange. `WP_Upgrader::$result` should never be `false` based on the property's documentation. 🤔

inc/packages/admin/info.php
afragen: Maybe `wp_kses_post()`?

inc/packages/class-upgrader.php
afragen:
```
		if ( ! $this->result || is_wp_error( $this->result ) ) {
			return $this->result;
		}
```

Implies it could be `false`

inc/packages/admin/info.php
afragen: Used the above if statement.

afragen:
I need to revert the changes to the coding-standards GitHub Action prior to merge — or not.

inc/packages/class-upgrader.php
costdev: Possibly, or an empty array which would be falsey. Not sure, but going with the Core method's docblock seems fine for now.

afragen: So resolved?

inc/packages/admin/namespace.php
```diff
@@ -151,38 +181,48 @@ function handle_direct_install() {
 	exit;
 }
 
+/**
+ * Hijack embedded inof page.
```
costdev:
```suggestion
 * Hijack embedded info page.
```

costdev (approved):
Think this is good for merge IMO. I'd say the remaining items (like debug/dump) can be handled before the main PR is merged.

PR #148, opened 2025-07-03
fix check_requirements and return bool

afragen:
Missed a call to check_requirements() and return boolean for the function.

### DIFF
--- a/inc/packages/admin/info.php
+++ b/inc/packages/admin/info.php
@@ -355,7 +355,7 @@ function version_requirements( ReleaseDocument $release ) {
  *
  * @param  ReleaseDocument $release Release document.
  *
- * @return void
+ * @return bool
  */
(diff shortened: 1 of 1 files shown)


inc/packages/admin/info.php
```diff
@@ -434,7 +436,7 @@ function get_action_button( MetadataDocument $doc, ReleaseDocument $release ) {
 	}
 
 	// Do we actually meet the requirements?
-	// $compatible = check_requirements( $release );
+	$compatible = check_requirements( $release );
 	$compatible = true;
```
costdev: Should we be removing the `$compatible = true;`?

afragen: You are correct.

costdev: Cool. Wasn't sure if we were forcing `true` for testing other parts at this stage.

PR #149, opened 2025-07-04
add check to update transient for default icons check

afragen:
I came across an error during another plugin update where the `$transient` did not have a `response` property. This fixes it.

### DIFF
--- a/inc/icons/namespace.php
+++ b/inc/icons/namespace.php
@@ -24,10 +24,12 @@ function bootstrap() {
  * @return stdClass
  */
 function set_default_icon( $transient ) {
-	foreach ( $transient->response as $updates ) {
-		$url = plugin_dir_url( PLUGIN_FILE ) . 'inc/icons/svg.php';
-		$url = add_query_arg( 'color', set_random_color(), $url );
-		$updates->icons['default'] = $url;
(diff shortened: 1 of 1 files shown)


inc/icons/namespace.php
```diff
@@ -24,6 +24,10 @@ function bootstrap() {
  * @return stdClass
  */
 function set_default_icon( $transient ) {
+	if ( ! property_exists( $transient, 'response' ) ) {
```
costdev, line 27: While we're here, should we also ensure that these are true?

1. `$transient->response` is a non-empty array.
    ```php
    if ( empty( $transient->response ) || ! is_array( $transient->response ) ) {
        return $transient;
    }
    ```
2. `$updates->icons` is set and is an array.

    ```php
    $url = plugin_dir_url...
    $url = add_query_arg...
    if ( ! isset( $updates->icons ) || ! is_array( $update->icons ) ) {
        $updates->icons = [];
    }
    $updates->icons['default'] = ...
    ```

afragen: If `$transient->response` is an empty array it will just run and exit.

`$updates->icons` is an array. And setting default data should be fine.

costdev (approved):

PR #150, opened 2025-07-04
Refactor for reuse of functions elsewhere

afragen:
Refactor allows for calling of init() elsewhere. Yes, I have a WIP for making slug-did show as true for `is_plugin_active(slug)`

### DIFF
--- a/inc/updater/namespace.php
+++ b/inc/updater/namespace.php
@@ -23,7 +23,7 @@ function bootstrap() {
  *
  * @return stdClass
  */
-function init() {
+function get_packages() {
 	/** @var array */
 	$packages = [];
 
@@ -67,7 +67,7 @@ function init() {
  * @return void
  */
 function run() {
-	$packages = init();
+	$packages = get_packages();
 	$plugins = $packages['plugins'] ?? [];
 	$themes = $packages['themes'] ?? [];
 	$packages = array_merge( $plugins, $themes);
--- a/inc/updater/namespace.php
+++ b/inc/updater/namespace.php
@@ -15,7 +15,7 @@ if ( ! defined( 'WPINC' ) ) {
  * Bootstrap.
  */
 function bootstrap() {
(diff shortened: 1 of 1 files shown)


inc/updater/namespace.php
```diff
@@ -54,20 +54,24 @@ function init() {
...
 }
 
 /**
  * Run Git Updater Lite for potential packages.
  *
  * @return void
  */
-function run( $package_arr ) {
-	foreach ( $package_arr as $package ) {
+function run() {
```
costdev, line 69: Is there any scenario where we may want to call `run( $packages )` like the function was before? Maybe for tests?

If so, it might be worth having:

```php
function bootstrap() {
    add_action( 'init', __NAMESPACE__ . '\\init' );
}

function init() {
    run( get_packages() );
}

function get_packages() {
    ...
}

function run( $packages ) {
    ...
}
```

afragen: Is that over complicating things? I'm having a difficult time understanding what you're trying to solve there.

costdev: I was just noticing that previously, `run()` could be called from anywhere using any packages, which may come in useful at another point during runtime, and having `get_packages()` and `run()` being independent of each other may be best.

Since the PR removes the parameter from `run()`, they're now tightly coupled, to each other and to `get_file_data()`, which also makes testing a little more awkward and can't independently verify each function as well as before.

afragen: What would help make it more testable?

costdev: Keeping an open mind and looking at it again. If there was more going on, then there might be more benefit in separating them. But there's not much going on in `run()` that someone couldn't achieve in a few lines.

Since this was more a thought and not a blocker, I think it's fine to keep as-is and if we want to open stuff up later or there's an issue in testing, we can revisit it.

costdev (approved):

PR #151, opened 2025-07-04
Add methods for slugs with DID

afragen:
This PR will/should make any installed FAIR plugin, that was installed with folder at `slug-did`, as an active plugin when using `is_plugin_active( slug )`.

Yes, I'm making a preference for avoiding name collisions by installing as `slug-did`.

Requires https://github.com/fairpm/fair-plugin/pull/150

### DIFF
--- a/inc/plugins/namespace.php
+++ b/inc/plugins/namespace.php
@@ -1,0 +1,103 @@
+<?php
+/**
+ * Sets DID-less plugin slug as active.
+ *
+ * @package FAIR
+ */
+
+namespace FAIR\Plugins;
+
+use function FAIR\Updater\init;
+
+/**
+ * Bootstrap
+ *
+ * @return void
+ */
+function bootstrap() {
+	add_filter( 'option_active_plugins', __NAMESPACE__ . '\\set_as_active' );
+	add_filter( 'wp_admin_notice_markup', __NAMESPACE__ . '\\hide_notice', 10, 3 );
+	if ( is_plugin_active( 'git-updater/git-updater.php' ) ) {
+		wp_admin_notice( 'Git Updater is active' );
+	}
+}
+
+/**
+ * Set FAIR plugins as active using DID-less slug.
+ *
+ * @param  array $active_plugins Array of active plugins.
+ *
+ * @return array
+ */
+function set_as_active( $active_plugins ) {
+	remove_filter( 'option_active_plugins', __NAMESPACE__ . '\\set_as_active' );
+	$packages = init();
+	$plugins = $packages['plugins'] ?? [];
+	foreach ( $plugins as $plugin ) {
+		if ( is_plugin_active( plugin_basename( $plugin ) ) ) {
+			$plugins[] = get_didless_slug( $plugin );
+		}
+	}
+	$active_plugins = array_map( 'plugin_basename', array_merge( $active_plugins, $plugins ) );
+
+	return array_unique( $active_plugins );
+}
+
+/**
+ * Return shortened DID withou `did:plc|web'.
+ *
+ * @param  string $did Full DID.
+ *
+ * @return string|void
+ */
+function get_short_did( $did ) {
+	if ( ! empty( $did ) ) {
+		if ( str_contains( $did, ':' ) ) {
+			$did = (string) explode( ':', $did )[2];
+		}
+		return $did;
+	}
+}
+
+/**
+ * Return slug without DID.
+ *
+ * @param  string $slug Current slug.
+ * @param  string $did  Full DID.
+ *
+ * @return string|void
+ */
+function get_didless_slug( $slug, $did = '' ) {
+	if ( empty( $did ) ) {
+		$did = get_file_data( $slug, [ 'PluginID' => 'Plugin ID' ] )['PluginID'];
+		$did = get_short_did( $did );
+	}
+	if ( ! empty( $did ) ) {
+		$slug = str_replace( '-' . get_short_did( $did ), '', $slug );
+	}
+
+	return $slug;
+}
+
+/**
+ * Hide notice reporting DID-less plugin is inactive because it doesn't exist.
+ *
+ * @param  string $markup Markup of notice.
+ * @param  string $message Message of notice.
+ * @param  array $args Args of notice.
+ *
+ * @return string
+ */
+function hide_notice( $markup, $message, $args ) {
+	$active = get_option( 'active_plugins' );
+	if ( $args['id'] === 'message' ) {
+		foreach ( $active as $plugin ) {
+			if ( str_contains( $message, $plugin ) && str_contains( $markup, 'error' ) ) {
+				remove_filter( 'wp_admin_notice_markup', __NAMESPACE__ . '\\hide_notice' );
+				return '';
+			}
+		}
+	}
+
+	return $markup;
+}
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -34,6 +34,7 @@ function bootstrap() {
 	Icons\bootstrap();
 	Importers\bootstrap();
 	Pings\bootstrap();
+	Plugins\bootstrap();
 	Salts\bootstrap();
 	Settings\bootstrap();
 	Updater\bootstrap();
--- a/plugin.php
+++ b/plugin.php
@@ -30,6 +30,7 @@ require_once __DIR__ . '/inc/disable-openverse/namespace.php';
 require_once __DIR__ . '/inc/icons/namespace.php';
 require_once __DIR__ . '/inc/importers/namespace.php';
 require_once __DIR__ . '/inc/pings/namespace.php';
+require_once __DIR__ . '/inc/plugins/namespace.php';
 require_once __DIR__ . '/inc/salts/namespace.php';
 require_once __DIR__ . '/inc/settings/namespace.php';
 require_once __DIR__ . '/inc/updater/namespace.php';


inc/plugins/namespace.php
```diff
@@ -0,0 +1,108 @@
...
+ * @param  string $markup Markup of notice.
+ * @param  string $message Message of notice.
+ * @param  array $args Args of notice.
+ *
+ * @return string
+ */
+function hide_notice( $markup, $message, $args ) {
+	$active = get_option( 'active_plugins' );
+	if ( $args['id'] === 'message' ) {
+		foreach ( $active as $plugin ) {
```
costdev:
```suggestion
	if ( $args['id'] === 'message' ) {
		$active = get_option( 'active_plugins' );
		foreach ( $active as $plugin ) {
```
No need to hit the database unless we know the ID is `'message'`.

inc/plugins/namespace.php
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/**
+ * Return shortened DID withou `did:plc|web'.
+ *
+ * @param  string $did Full DID.
+ *
+ * @return string|void
+ */
+function get_short_did( $did ) {
```
costdev: The "short DID" as referred to here is known as the "method-specific-id" in the spec. I think we should try to stick to the terminology where possible. Besides "short" is vague :wink:. What about `get_did_id()`?

inc/plugins/namespace.php
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * @return array
+ */
+function set_as_active( $active_plugins ) {
+	remove_filter( 'option_active_plugins', __NAMESPACE__ . '\\set_as_active' );
+	$packages = get_packages();
+	$plugins = $packages['plugins'] ?? [];
+	foreach ( $plugins as $plugin ) {
+		if ( is_plugin_active( plugin_basename( $plugin ) ) ) {
+			$plugins[] = get_didless_slug( $plugin );
```
costdev: Careful - `$plugin` isn't a slug in this situation, it's `trailingslashit( WP_PLUGIN_DIR ) . $file` - [Ref](https://github.com/fairpm/fair-plugin/blob/main/inc%2Fupdater%2Fnamespace.php#L44).

inc/plugins/namespace.php
costdev: Aside from that, I think there's a great opportunity here for a more general function, `parse_did()` that returns an array keyed on `'method'` and `'id'`, or `false` for an improper DID.

So instead of having `get_did_id()`, you'd instead have:

```php
function parse_did( $did ) {
  $parts = explode( ':', $did, 3 );

  if (
    count( $parts ) !== 3 ||
    $parts[0] !== 'did' ||
    ( $parts[1] !== 'web' && $parts[1] !== 'plc' )
  ) {
    return false;
  }

  return [
    'method' => $parts[1],
    'id' => $parts[2],
  ];
}
```

Usage:
```php
$did = parse_did( $did );

if ( ! $did ) {
  // Error or something else.
}

$id = $did['id'];
```

Actually, there's already a `parse_did()` being introduced in #71 - [here's its implementation](https://github.com/fairpm/fair-plugin/blob/add-fair-installer/inc%2Fpackages%2Fnamespace.php#L35-L55).

Seems like we'll end up using that function when #71 gets merged.

inc/plugins/namespace.php
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Return shortened DID withou `did:plc|web'.
+ *
+ * @param  string $did Full DID.
+ *
+ * @return string|void
+ */
+function get_short_did( $did ) {
+	if ( ! empty( $did ) ) {
+		if ( str_contains( $did, ':' ) ) {
```
costdev: I'll be looking at what our minimum WP version should be bumped to, and opening a PR as soon as I can, so this doesn't need to be changed to `strpos()`.

Still, with a current minimum PHP of 7.4 and WP of 5.4, bad Andy for using this function. *hits with ruler*

costdev (changes requested):
Some changes, some thoughts, and a whack with a ruler. This review has everything!

inc/plugins/namespace.php
afragen: Except that when I didn't have `plugin_basename()` it didn't see as active. I'll test again.

costdev: Oh, I meant the call to `get_didless_slug( $plugin )`, which expects the argument to be a slug, not a path. Maybe it's `get_didless_slug()`'s internally referring to `$slug` that's causing confusion for me. 

Would it be clearer to rename `get_didless_slug()` to `remove_did_id()` and call the parameter `$plugin`? That way, `$plugin` could be a slug, an absolute path, or `slug/file.php`.

inc/plugins/namespace.php
afragen: str_contains() is in a shim in WP5.9

costdev: Yep, and our current minimum is 5.4. We also use `wp_admin_notice()`, which is WP 6.4, so I know we'll need to at least bump our minimum to 6.4. I'll work out if there are any other uses that need us to bump above 6.4.

afragen: Besides Ryan uses `str_starts_with()` in `parse_did()` 😆

inc/plugins/namespace.php
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Return slug without DID.
+ *
+ * @param  string $slug Current slug.
+ * @param  string $did  Full DID.
+ *
+ * @return string|void
+ */
+function get_didless_slug( $slug, $did = '' ) {
```
Ipstenu: Hm. If we do `get_did_id()` then this might be better as `get_slug_without_did_id()`?

inc/plugins/namespace.php
Ipstenu: Thank you @costdev !

inc/plugins/namespace.php
afragen: Refactored in https://github.com/fairpm/fair-plugin/pull/151/commits/5795d008f1be7f7d9a4f35a4c3fc87cce784dc0e

inc/plugins/namespace.php
afragen: Refactored in https://github.com/fairpm/fair-plugin/pull/151/commits/5795d008f1be7f7d9a4f35a4c3fc87cce784dc0e

inc/plugins/namespace.php
afragen: Good idea, I sort of did that in https://github.com/fairpm/fair-plugin/pull/151/commits/5795d008f1be7f7d9a4f35a4c3fc87cce784dc0e

inc/plugins/namespace.php
```diff
@@ -0,0 +1,116 @@
...
+	add_filter( 'option_active_plugins', __NAMESPACE__ . '\\set_as_active' );
+	add_filter( 'wp_admin_notice_markup', __NAMESPACE__ . '\\hide_notice', 10, 3 );
+
+	// just for testing.
+	if ( ! function_exists( 'is_plugin_active' ) ) {
+		require_once ABSPATH . 'wp-admin/includes/plugin.php';
+	}
+	if ( is_plugin_active( 'git-updater/git-updater.php' ) ) {
+		wp_admin_notice( 'Git Updater is active' );
+	}
```
afragen, line 27: We will need to remove this when ready to merge.

afragen:
I have a better way of getting the method specific ID from the DID. I plan on extending Ryan's `parse_did()`. This will change a number of things in this PR, hopefully simplifying it. It also means the PR to extend `parse_did()` will need to be evaluated first. See #152

afragen:
~~Needs #153 to pass tests.~~

inc/plugins/namespace.php
```diff
@@ -0,0 +1,93 @@
...
+function get_file_without_did_hash( $plugin ) : string {
+	$file = explode( '/', $plugin, 2 );
+	$slug_parts = explode( '-', $file[0] );
+
+	// Remove hash.
+	array_pop( $slug_parts );
+	$slug = implode( '-', $slug_parts );
+
+	return $slug . '/' . $file[1];
+}
```
costdev, line 76: Can save the explode/pop/implode with this:

```php
function get_file_without_did_hash( $plugin ) : string { 
  list( $slug, $file ) = explode( '/', $plugin, 2 ); 
  $slug = substr( $slug, 0, strrpos( $slug, '-' ) );

  return "{$slug}/{$file}"; 
}
```

Or just based on the known length:

```php
function get_file_without_did_hash( $plugin ) : string { 
  list( $slug, $file ) = explode( '/', $plugin, 2 );

  return substr( $slug, 0, -7 ) . "/{$file}";
}
```

It's fine as-is too, I'm just dropping thoughts

inc/plugins/namespace.php
```diff
@@ -0,0 +1,93 @@
...
+ * @return array
+ */
+function set_as_active( $active_plugins ) {
+	remove_filter( 'option_active_plugins', __NAMESPACE__ . '\\set_as_active' );
+	$packages = get_packages();
+	$plugins = $packages['plugins'] ?? [];
+	$plugins = array_map( 'plugin_basename', $plugins );
+	foreach ( $plugins as $plugin ) {
+		if ( is_plugin_active( $plugin ) ) {
+			$plugins[] = get_file_without_did_hash( $plugin );
```
costdev: Instead of adding this to `$plugins`, then doing `array_merge()` after the loop, couldn't this be added directly to `$active_plugins`? The `array_unique()` is fine of course since it saves on in-loop `in_array()` checks / `array_flip()` flopping.

inc/plugins/namespace.php
```diff
@@ -0,0 +1,93 @@
...
+ * @param  string $message Message of notice.
+ * @param  array $args Args of notice.
+ *
+ * @return string
+ */
+function hide_notice( $markup, $message, $args ) {
+	if ( $args['id'] === 'message' ) {
+		$active = get_option( 'active_plugins' );
+		foreach ( $active as $plugin ) {
+			if ( str_contains( $message, $plugin ) && str_contains( $markup, 'error' ) ) {
```
costdev, line 91: Check for the `'error'` before the loop, in the initial `if` - can alternatively use `in_array( 'error', $args['additional_classes'], true )` instead of checking for `'error'` somewhere in the markup.

inc/plugins/namespace.php
```diff
@@ -0,0 +1,93 @@
...
+use function FAIR\Updater\get_packages;
+
+/**
+ * Bootstrap
+ *
+ * @return void
+ */
+function bootstrap() {
+	add_filter( 'option_active_plugins', __NAMESPACE__ . '\\set_as_active' );
+	add_filter( 'wp_admin_notice_markup', __NAMESPACE__ . '\\hide_notice', 10, 3 );
```
costdev: Suggest only hooking this on `load-plugins.php` and such rather than hooking on every notice on every page.

inc/plugins/namespace.php
afragen: Probably could be.

inc/plugins/namespace.php
afragen: But then I need to also check that the `additional_classes` is set correctly.

inc/plugins/namespace.php
costdev: Worth a try. I'm on mobile and the battery's on 1 perc- ... ... Oh, I'm still here. So as I was saying, I'm on mobile otherwise I'd test it myse-

inc/plugins/namespace.php
afragen: Does this actually work for `my-super-awesome-wonderful-plugin-ju94ks/plugin.php`?

afragen:
This will error until #71 is merged. It requires `get_did_hash()`.

inc/plugins/namespace.php
afragen: Are you certain that's where we want to limit it?

inc/plugins/namespace.php
costdev: Yeah both work for that: [3v4l](https://3v4l.org/c8aE8#vnull)

inc/plugins/namespace.php
costdev: Can just use:
```php
in_array( 'error', $args['additional_classes'] ?? [], true )
```

costdev: I'm noticing that there's [another error](https://github.com/WordPress/wordpress-develop/blob/6bc94efc15e0be4ebfd9a209cfd2f1b65bab8f7c/src/wp-admin/plugins.php#L719) that has an ID of `message` and has the `error` class.

I think, unfortunately, we'll have to check for `__( 'Plugin file does not exist.' )` and add a `phpcs:ignore` comment for the lack of textdomain since this is a Core string.

inc/plugins/namespace.php
costdev: Yeah that's the only file where the notice appears, where the deactivation error string is used, and the only place in Core where `validate_active_plugins()` or `validate_plugin()` are called (other than `activate_plugin()`).

Not being able to filter those functions is a real shame.

inc/plugins/namespace.php
costdev: Sneaky!

In that case, couldn't we just do this?

```php
function get_file_without_did_hash( $did, $plugin ) : string {
  list( $slug, $file ) = explode( '/', $plugin, 2 );
  $slug = str_replace( '-' . get_did_hash( $did ), '', $slug );

  return $slug . '/' . $file;
}
```

costdev: That way, we're removing the hash if it's there, otherwise we're basically just returning what was passed, which is already a hashless slug.

inc/plugins/namespace.php
afragen: The markup won't have the `__()`.

inc/plugins/namespace.php
```diff
@@ -0,0 +1,99 @@
...
+	add_action( 'load-plugins.php', __NAMESPACE__ . '\\load_filters' );
+}
+
+/**
+ * Load filters.
+ *
+ * @return void
+ */
+function load_filters() {
+	add_filter( 'option_active_plugins', __NAMESPACE__ . '\\set_as_active' );
```
costdev, line 36: I think this active plugins hook should still be in the bootstrap since that option is accessed in many more places, whereas the admin notice is only on `plugins.php`

rmccue:
I will say that I don't love the pattern of hook-and-self-unhook here, and it also doesn't seem like it works consistently - what happens when you call `is_plugin_active()` twice? Seems like the code here breaks with that. Additionally, filtering `option_active_plugins` on _every_ call seems like it'd have the potential to accidentally get saved into the DB, which is dangerous.

I'd rather we instead a) install consistently into `{slug}-{hash}/` meaning that code can rely on that pretty consistently, and b) encourage the use of `is_plugin_active( 'did:...' )` instead of the filenames anyway.

afragen:
@rmccue consider this more of a POC. It does seem that your hijack and backtrace method might be cleaner.

PR #152, opened 2025-07-06
Get and return method specific ID with parse_did()

afragen:
Extends `FAIR\Packages\parse_did()` to also return the method specific ID from the full DID.

If you like something better than `$short_id`, LMK.

### DIFF
--- a/inc/packages/did/class-plc.php
+++ b/inc/packages/did/class-plc.php
@@ -30,17 +30,17 @@ class PLC implements DID {
 	 *
 	 * @var string
 	 */
-	protected string $short_id;
+	protected string $msid;
 
 	/**
 	 * Constructor.
 	 *
 	 * @param string $id DID.
-	 * @param string $short_id Method specific ID.
+	 * @param string $msid Method specific ID.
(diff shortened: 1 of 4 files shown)


costdev:
I think it may be worth changing to use `::get()` for the full `did:method:method-specific-id`, `::get_method()` for `method`, and `get_id()` for `id` (since each class is for a specific method anyway).

It doesn't seem worth changing terminology and having to add `short_id()` because `id` is taken.

inc/packages/did/class-plc.php
```diff
@@ -50,6 +59,13 @@ public function get_id() : string {
 		return $this->id;
 	}
 
+	/**
+	 * Get the method specific ID from DID.
+	 */
+	public function get_msid() : string {
```
Ipstenu, line 65: I can't believe I'm saying this but msid is known and used in Microsoft dev work.

`get_method_specific_id()` would be better for clarity anyway. It's not like we _need_ to shorthand everything

afragen:
Perhaps we should close this one?

PR #154, opened 2025-07-08
Add DID hash function

### LINKED ISSUE #124: Deduplication
This issue, stemming from comments on #115, is for discussing what deduplication is, why it's needed, what the problems are, and how FAIR might overcome them. I'd like to make sure that we discuss and consider all of the aspects involved before we consider implementing deduplication.

Some of these aspects and considerations are included below to get us started.

### Why deduplication is needed
WordPress uses the slug as the directory name where a plugin or theme is installed. A plugin and a theme can have the same slug. However, when **two plugins** or **two themes** exist with the same slug, this results in a conflict where only one can be installed at a time.

### How deduplication is handled elsewhere
- #### Installation
  For sites like WordPress.org, a slug that is already used by another plugin/theme is not allowed, preventing duplicates on WordPress.org. However, this doesn't prevent duplicates for plugins or themes not hosted on WordPress.org.

- #### Updating
  An `Update URI` header exists for plugins <sup>[[1]](https://developer.wordpress.org/plugins/plugin-basics/header-requirements/#header-fields)</sup> and themes  <sup>[[2]](https://developer.wordpress.org/themes/basics/main-stylesheet-style-css/#explanations)</sup> so that the developer can specify that it is not updated from WordPress.org.

### Why FAIR should try to solve this existing problem
While this is an existing problem in the ecosystem, it becomes more visible with a federated system which makes non-WordPress.org sources easily discoverable to users.

### Some possible approaches
- #### Renaming the directory
  So far, one suggestion has been to rename a plugin's or theme's directory to something unique prevent a conflict. For example, instead of using `slug`, it could use `slug-did`.

- #### Other approaches
  Add your thoughts on alternative approaches in this issue.

### Considerations

- #### Native WordPress functions and hooks
  Functions such as [wp_get_theme](https://developer.wordpress.org/reference/functions/wp_get_theme/), [is_theme_paused](https://developer.wordpress.org/reference/functions/is_theme_paused/), [wp_is_theme_directory_ignored](https://developer.wordpress.org/reference/functions/wp_is_theme_directory_ignored/), [is_plugin_active](https://developer.wordpress.org/reference/functions/is_plugin_active/), [is_plugin_inactive](https://developer.wordpress.org/reference/functions/is_plugin_inactive/), [is_plugin_paused](https://developer.wordpress.org/reference/functions/is_plugin_paused/), [is_network_only_plugin](https://developer.wordpress.org/reference/functions/is_network_only_plugin/), [is_plugin_active_for_network](https://developer.wordpress.org/reference/functions/is_plugin_active_for_network/) and others, as well as various hooks receive an argument either in the `slug` or `slug/file.php` formats. Many if not all such functions **are not pluggable**. Many plugins and themes, including site-specific code make use of such functions. If the directory is simply renamed with no further action, these functions will give false results.

  At first, it may seem straightforward enough for an individual developer to change their calls or add a second call to these functions/hooks which includes the `slug` or `slug-did/file.php` format as needed. However, the need for deduplication means that `slug` or `slug/file.php` could exist and would lead to false results. Since not all sites will be using FAIR, a developer couldn't just rely on `slug-did` or `slug-did/file.php` either.

  Many of functions are used internally by WordPress Core, and requests to implement checks for specific plugins are generally rejected. However, it may be possible to request that any of these functions which are not currently pluggable are made pluggable.

- #### Getting adoption for custom functions
  If we were to provide alternatives for these functions, this would require adoption by every maintained plugin, theme, and site-specific code that uses the native WordPress functions, as well as tutorials which influence many developer decisions. As we know, this is a near impossible task.

## Your thoughts
I'd like to gather input, not only on the above, but any other aspects or considerations you have in mind. If you have opinions, thoughts, ideas, proofs of concept, or anything else to contribute, drop a comment below and let's discuss!

afragen:
Assuming we've decided on a format of `$hash = substr( hash( 'sha256', $did->get_id() ), 0, 6 );`. This creates a simple function to get that hash.

Fixes #124

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -55,6 +55,19 @@ function parse_did( string $id ) {
 }
 
 /**
+ * Return hash of DID for appending to slug.
+ *
+ * @param  string $id DID
+ *
+ * @return string
+ */
+function get_did_hash( string $id ) : string {
+	$did = parse_did( $id );
+
+	return substr( hash( 'sha256', $did->get_id() ), 0, 6 );
+}
+
+/**
  * Get DID document.
  *
  * @param string $id DID.
--- a/inc/packages/class-upgrader.php
+++ b/inc/packages/class-upgrader.php
@@ -675,7 +675,7 @@ class Upgrader extends WP_Upgrader {
 			return $source;
 		}
 
-		$new_source = trailingslashit( $remote_source ) . $this->package->slug;
+		$new_source = trailingslashit( $remote_source ) . $this->package->slug . '-' . get_did_hash( $this->package->id );
 
 		if ( trailingslashit( strtolower( $source ) ) !== trailingslashit( strtolower( $new_source ) ) ) {
 			$wp_filesystem->move( $source, $new_source, true );


inc/packages/namespace.php
```diff
@@ -54,6 +54,19 @@ function parse_did( string $id ) {
...
 
+/**
+ * Return hash of DID for appending to slug.
+ *
+ * @param  string $id DID
+ *
+ * @return string
+ */
+function get_did_hash( string $id ) : string {
+	$did = parse_did( $id );
```
costdev, line 65: `parse_did()` can return a `WP_Error` object too. Should check for this here right?

inc/packages/namespace.php
```diff
@@ -54,6 +54,19 @@ function parse_did( string $id ) {
 	}
 }
 
+/**
+ * Return hash of DID for appending to slug.
```
costdev, line 58: Just a suggestion: Rather than tie the function to this initial use case in its summary, what about keeping the summary to its behaviour, and outlining the general benefit of the function in its description?

Something like this:
```php
/**
 * Hash a DID.
 *
 * This ensures a consistent representation of a DID
 * regardless of its method, method-specific-id format, or length.
 *
 * ...
 */
```

PR #156, opened 2025-07-09
Pass at updating via DID

afragen:
Needs https://github.com/fairpm/mini-fair-repo/pull/19 for `mini-fair-repo`.

Whole bunch of stuff other than start updating installed FAIR plugins via DID. It actually works, mostly. I haven't been able to test themes until there's one to test with.

screenshot_678

screenshot_679

Yes, missing the button.

screenshot_680

Yes, it does work.

Sorry, for some reason `composer format` doesn't seem to be working locally for me.

### DIFF
--- a/inc/packages/class-metadatadocument.php
+++ b/inc/packages/class-metadatadocument.php
@@ -43,6 +43,13 @@ class MetadataDocument {
 	public $slug;
 
 	/**
+	 * File name.
+	 *
+	 * @var string
+	 */
+	public $file;
+
+	/**
 	 * License.
 	 *
 	 * @var string
--- a/inc/packages/class-metadatadocument.php
+++ b/inc/packages/class-metadatadocument.php
@@ -124,6 +124,7 @@ class MetadataDocument {
 		$optional = [
 			'name',
 			'slug',
+			'file',
 			'description',
 			'keywords',
 			'sections',
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -74,6 +74,21 @@ function get_did_hash( string $id ) {
 }
 
 /**
+ * Get file with DID hash.
+ *
+ * @param string $did DID.
+ * @param string $file File name.
+ *
+ * @return string
+ */
+function get_file_with_did_hash( $did, $file ) {
+	list( $slug, $file ) = explode( '/', $file, 2 );
+	$slug = $slug . '-' . get_did_hash( $did );
+
+	return $slug . '/' . $file;
+}
+
+/**
  * Get DID document.
  *
  * @param string $id DID.
--- a/inc/updater/class-fair-updater.php
+++ b/inc/updater/class-fair-updater.php
@@ -45,11 +45,7 @@ class FAIR_Updater {
 	/** @var string */
 	protected $type;
 
-	/**
-	 * Release document.
-	 *
-	 * @var ReleaseDocument
-	 */
+	/** @var ReleaseDocument */
 	protected $release;
 
 	// phpcs:enable
--- a/inc/updater/class-fair-updater.php
+++ b/inc/updater/class-fair-updater.php
@@ -23,6 +23,7 @@ use function FAIR\Packages\fetch_metadata_doc;
 use function FAIR\Packages\fetch_package_metadata;
 use function FAIR\Packages\get_did_document;
 use function FAIR\Packages\get_did_hash;
+use function FAIR\Packages\get_file_with_did_hash;
 use function FAIR\Packages\pick_release;
 
 /**
--- a/inc/updater/class-fair-updater.php
+++ b/inc/updater/class-fair-updater.php
@@ -1,0 +1,462 @@
+<?php
+/*
+ * @author   Andy Fragen
+ * @license  MIT
+ * @link     https://github.com/afragen/git-updater-lite
+ * @package  git-updater-lite
+ */
+
+namespace FAIR\Updater;
+
+use FAIR\Packages\MetadataDocument;
+use FAIR\Packages\ReleaseDocument;
+use Plugin_Upgrader;
+use stdClass;
+use Theme_Upgrader;
+use TypeError;
+use WP_Error;
+use WP_Upgrader;
+
+use const FAIR\Packages\SERVICE_ID;
+
+use function FAIR\Packages\fetch_metadata_doc;
+use function FAIR\Packages\fetch_package_metadata;
+use function FAIR\Packages\get_did_document;
+use function FAIR\Packages\get_did_hash;
+use function FAIR\Packages\pick_release;
+
+/**
+ * Class FAIR_Updater.
+ */
+class FAIR_Updater {
+
+	// phpcs:disable Generic.Commenting.DocComment.MissingShort
+
+	/** @var string */
+	protected $local_version;
+
+	/** @var string */
+	protected $did;
+
+	/** @var MetadataDocument */
+	protected $metadata;
+
+	/** @var string */
+	protected $type;
+
+	/**
+	 * Release document.
+	 *
+	 * @var ReleaseDocument
+	 */
+	protected $release;
+
+	// phpcs:enable
+
+	/**
+	 * Constructor.
+	 *
+	 * @param array $packages Associative array as `'DID' => Path to package`.
+	 */
+	public function __construct( string $did, string $filepath ) {
+		$this->did = $did;
+		$data  = get_file_data( $filepath, array( 'Version'   => 'Version'));
+		$this->local_version = $data['Version'];
+	}
+
+	/**
+	 * Get the required versions, (PHP, WP, Tested to).
+	 *
+	 * @return array
+	 */
+	protected function get_required_versions() {
+		foreach ( $this->release->requires as $pkg => $vers ){
+			$vers = preg_replace( '/^[^0-9]+/', '', $vers );
+		if ( $pkg === 'env:php' ) {
+			$required_versions['requires_php'] = $vers;
+		}
+		if ( $pkg === 'env:wp' ) {
+			$required_versions['requires_wp'] = $vers;
+		}
+		}
+		foreach ( $this->release->suggests as $pkg => $vers ) {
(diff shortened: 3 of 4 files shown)


afragen:
screenshot_681

I found the button.

inc/packages/namespace.php
```diff
@@ -73,6 +73,21 @@ function get_did_hash( string $id ) {
...
+ * Get file with DID hash.
+ *
+ * @param string $did DID.
+ * @param string $file File name.
+ *
+ * @return string
+ */
+function get_file_with_did_hash( $did, $file ) {
+	list( $slug, $file ) = explode( '/', $file, 2 );
+	$slug = $slug . '-' . get_did_hash( $did );
```
costdev: Suggest checking if the slug ends with the hash before adding it. Saves having to check before calling this function, and means it can be used just to be sure the hash is there.

```php
$did_hash = get_did_hash( $did );
if ( ! str_ends_with( $slug, $did_hash ) ) {
  $slug .= '-' . $did_hash;
}
```

inc/updater/class-fair-updater.php
```diff
@@ -0,0 +1,460 @@
...
+		$this->local_version = $data['Version'];
+	}
+
+	/**
+	 * Get the required versions, (PHP, WP, Tested to).
+	 *
+	 * @return array
+	 */
+	protected function get_required_versions() {
+		foreach ( $this->release->requires as $pkg => $vers ){
```
costdev:
```suggestion
		$required_versions = [];
		foreach ( $this->release->requires as $pkg => $vers ){
```
`$required_versions` doesn't seem to be initialized. Even though [autokittenvacation on false](https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.core.autovivification-false) was deprecated in PHP 8.1, not on `null`, might as well just initialize it to the empty array for clarity and, you know, "just in case" `null` gets deprecated later.

inc/updater/class-fair-updater.php
```diff
@@ -0,0 +1,460 @@
...
+		return new WP_Error( 'fair.packages.install_plugin.no_signing_keys', __( 'DID does not contain valid signing keys.', 'fair' ) );
+	}
+
+	$metadata = fetch_metadata_doc( $repo_url );
+	if ( is_wp_error( $metadata ) ) {
+		return $metadata;
+	}
+
+	// Select the latest release.
+	$releases = array_values( $metadata->releases );
```
costdev: Curious, why `array_values()` on this?

inc/updater/namespace.php
```diff
@@ -35,32 +31,93 @@ function get_packages() {
...
+ */
+function get_icons($icon) {
+	if ( empty($icon)){
+		return;
+	}
+
+	$icons_arr = [];
+	$icons = $icon;
+
+	$regular = array_find( $icons, fn ( $icon ) => $icon->width === 772 && $icon->height === 250 );
```
costdev, line 124: Note: Using `array_find()` will bump our minimum WP version to 6.8, which is when [the polyfill for PHP 8.4's array_find()](https://core.trac.wordpress.org/ticket/62558) was introduced.

inc/updater/namespace.php
```diff
@@ -35,32 +31,93 @@ function get_packages() {
...
+ *
+ * @return array
+ */
+function get_banners( $banner ){
+	if ( empty( $banner ) ) {
+		return;
+	}
+
+	$banners_arr = [];
+	$banners = $banner;
```
costdev: Any reason not to just rename the parameter to `$banners`?

inc/updater/class-fair-updater.php
```diff
@@ -0,0 +1,461 @@
...
+use function FAIR\Packages\fetch_package_metadata;
+use function FAIR\Packages\get_did_document;
+use function FAIR\Packages\get_did_hash;
+use function FAIR\Packages\get_file_with_did_hash;
+use function FAIR\Packages\pick_release;
+
+/**
+ * Class FAIR_Updater.
+ */
+class FAIR_Updater {
```
costdev: We don't need to prefix with `FAIR_` when we're already namespaced. If that's just preference, no worries. Just noting it because I saw it.

costdev (commented):
There are some indentation issues and some inconsistent spacing between parentheses. Didn't flag each instance, but PHPCBF should be able to autofix those.

afragen:
PHPCBF not working, at least not locally for me.

inc/updater/class-fair-updater.php
afragen: Copied from Ryan's code elsewhere.

costdev: @rmccue Tag, you're it. No issues, I'm just curious why `array_values()` is used here.

inc/updater/class-fair-updater.php
afragen: It needs to match the file name.

inc/updater/namespace.php
afragen: Yes, because $banner and $banners are both used.

costdev: `$banner` only appears to be used in the `array_find()` callback, which in that scope is the parameter in `fn ( $banner )`, right?

inc/updater/namespace.php
```diff
@@ -35,32 +35,92 @@ function get_packages() {
...
+/**
+ * Get banners.
+ *
+ * @param  array $banner Banner data.
+ *
+ * @return array
+ */
+function get_banners( $banner ) {
+	if ( empty( $banner ) ) {
+		return;
```
costdev:
```suggestion
		return [];
```
Docblock says it should return an array.

inc/updater/namespace.php
afragen: Again, I copied Ryan's code. Which works.

costdev: That's fine, we just need to make sure we either bump the minimum WP version to 6.8, or we add the polyfill behind a `function_exists()` guard.

inc/updater/namespace.php
afragen: Again, Ryan's code. Doesn't make it right, but it does mean I copied it.

inc/updater/class-fair-updater.php
afragen: File renamed.

afragen:
To test you'll have to set the feature flag as in the README for #71. You will need the current code in #71 as well as the current code here.

Then install Git Updater via the DID (did:plc:afjf7gsjzsqmgc7dlhb553mv). Locally decrease the version number and test away.

inc/updater/namespace.php
afragen: Polyfill might be a better idea. Especially if we want to require a lower WP version.

inc/updater/namespace.php
afragen: Updated. Colin is right.

afragen:
Setting the `slug` in the `get_update_data()` to `slug-hash` fixes the button in View details modal.

screenshot_682

inc/packages/class-metadatadocument.php
```diff
@@ -42,6 +42,20 @@ class MetadataDocument {
...
+	 * @var string
+	 */
+	public $file;
+
+	/**
+	 * Is package binary a release asset?
+	 *
+	 * @var bool
+	 */
+	public $release_asset;
```
costdev: I'm curious whether we need this additional property, or if we should actually just be using the `$releases` property which contains an array of `ReleaseDocument` objects, which I assume means they're release assets.

Is there some way to use what's in that property to accomplish the goal here?

afragen: There is no identifier of a release asset. Closest thing would be to parse download URI but that's not a good idea.

afragen: GitHub release assets require an Accept header of application/octet-stream in order to download.

costdev: I suppose I'm wondering this:

If we have a `MetadataDocument` class and a `ReleaseDocument` class, why can a `MetadataDocument` object be considered a release asset? Shouldn't the only release assets be `ReleaseDocument` objects?

afragen: A release asset is a type of package binary.

afragen: Perhaps a better method of detection using, https://github.com/fairpm/mini-fair-repo/pull/21

inc/packages/namespace.php
```diff
@@ -73,6 +73,21 @@ function get_did_hash( string $id ) {
...
+ * Get file with DID hash.
+ *
+ * @param string $did DID.
+ * @param string $file File name.
+ *
+ * @return string
+ */
+function get_file_with_did_hash( $did, $file ) {
+	list( $slug, $file ) = explode( '/', $file, 2 );
+	$slug .= '-' . get_did_hash( $did );
```
costdev: Should we have a check here in case it already ends with it?

Like:
```php
$did_hash = get_did_hash();
if ( ! str_ends_with( $did, '-' . $did_hash ) ) {
  $slug .= '-' . $did_hash;
}
```

inc/updater/namespace.php
rmccue: Yeah, we should copy across the polyfill; we can handle that in a separate PR though.

inc/packages/namespace.php
afragen: The use case here is adding the hash to the non-hashed filename.

costdev: Sure, I was just thinking maybe someone would call this function without realising a filename may already have it. Anyway, it's not necessary, just a thought.

afragen: It's only used in one place and we might be replacing that use case. I'll remove the function and just put the code where the use case is.

inc/updater/namespace.php
afragen: https://github.com/fairpm/fair-plugin/issues/163

Ipstenu (approved):

PR #160, opened 2025-07-12
Ping IndexNow for newly 404ing URLs.

### LINKED ISSUE #158: Ping IndexNow when old status is published.
The IndexNow protocol allows for pinging the service with URLs that 404 or 410 response. [From their FAQ](https://www.indexnow.org/faq):

> Can I submit 404 URLs through the API?
>
> Yes, you can submit dead links (http 404, http 410) pages to notify search engines about new dead links.

This plugin currently only sends pings for post object with the publish status, I think it would be helpful to change the condition to ping if either the new or old status is publish. 

https://github.com/fairpm/fair-plugin/blob/ffbc7559a7fabcc2bc8f2048786107ac4ebd17bb/inc/pings/namespace.php#L112-L115

I can work on a PR for this.

peterwilsoncc:
Modifies the conditions under which IndexNow is pinged.

**Why**

1. To ensure that newly 404ing URLs are sent to hosts to encourage them to de-index them. 
   In the event someone has accidentally published a page and then unpublished it, it's likely they wish to have the URL removed from search engines asap.
2. Prevent double pings for posts on sites that use legacy meta boxes (a very popular SEO plugin still uses them so it's not uncommon)

**How**

Replaces the check for `'publish' !== $new_status` with a check whether the old and new post statuses are publicly viewable.

If both are not viewable no ping takes place as the URL will be an existing 404 response.

If one is viewable but the other is not, a ping takes place as the page has either being updated or unpublished.

As sites can have custom public post statuses, I've removed the hard coded `publish` status check as `is_post_status_viewable()` will cover a wider range of statuses, including `publish`.

No ping takes place for non-viewable post types as the it shouldn't change without developer intervention, at which point they are responsible for the SEO impacts of the change.

For item 2: Bypass pinging if `'1' == $_GET['meta-box-loader']`

Fixes #158.

### DIFF
--- a/inc/pings/namespace.php
+++ b/inc/pings/namespace.php
@@ -109,18 +109,51 @@ function handle_key_file_request() {
  * @param WP_Post $post       Post object.
  */
 function ping_indexnow( $new_status, $old_status, $post ) : void {
-	// Only ping for published posts.
-	if ( 'publish' !== $new_status ) {
+	/*
+	 * Skip if post type isn't viewable.
+	 *
+	 * The post type shouldn't change under normal circumstances,
+	 * so it's safe to assume that both the old and new post are
+	 * not viewable.
+	 */
+	if ( ! is_post_type_viewable( $post->post_type ) ) {
 		return;
 	}
 
-	// Skip revisions and autosaves.
+	/*
+	 * Skip for revisions and autosaves.
+	 *
+	 * The IndexNow ping for revisions will be handled by the
+	 * parent post's transition_post_status hook.
+	 */
 	if ( wp_is_post_revision( $post ) || wp_is_post_autosave( $post ) ) {
 		return;
 	}
 
-	// Skip non-public post types.
-	if ( ! is_post_type_viewable( $post->post_type ) ) {
+	/*
+	 * Skip if both old an new statuses are private.
+	 *
+	 * The page will have been a 404 before and after.
+	 *
+	 * For pages that are newly a 404, still ping IndexNow
+	 * to encourage removal of the URL from search engines.
+	 */
+	if (
+		! is_post_status_viewable( $new_status )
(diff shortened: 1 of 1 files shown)


inc/pings/namespace.php
```diff
@@ -109,18 +109,51 @@ function handle_key_file_request() {
...
+	 * parent post's transition_post_status hook.
+	 */
 	if ( wp_is_post_revision( $post ) || wp_is_post_autosave( $post ) ) {
 		return;
 	}
 
-	// Skip non-public post types.
-	if ( ! is_post_type_viewable( $post->post_type ) ) {
+	/*
+	 * Skip if both old an new statuses are private.
```
rmccue:
```suggestion
	 * Skip if both old and new statuses are private.
```

rmccue (approved):
This looks good to me; would appreciate @jdevalk's eyes on as well.

jdevalk (approved):
LGTM!

For reference, as I did just verify, the FAQ on https://www.indexnow.org/faq explicitly states that pinging with newly 404ing URLs to get them removed is fine, so this is a solid improvement.

peterwilsoncc:
I've checked the PHPCS failures, they're happening in an unrelated file.

FWIW, in the implementation for my site (very heavily based on this code), I also removed the revision and autosave check to allow `is_post_status_viewable()` handle the bypass but I figured it was safer to leave them in place for this code.

peterwilsoncc:
@rmccue @jdevalk @afragen 

I got this wrong. The newly 404ing URLs don't use the correct permalink as it returns the plain version, ie example.com/?p=1234.

I'm figuring it all out in a plugin to account for various use cases. Including tests. https://github.com/peterwilsoncc/no-fuss-indexnow

PR #164, opened 2025-07-15
Add WP Core compat.php

### LINKED ISSUE #163: Add WP polyfills for backwards PHP compatibility
Given the many advanced PHP functions used we will need to include the WP core `compat.php` file to maintain some sense of backwards compatibility with older PHP versions.

https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/compat.php

afragen:
Add WP Core's compat.php for backwards PHP compatibility.

Fixes #163

### DIFF
--- a/inc/compatibility/compat.php
+++ b/inc/compatibility/compat.php
@@ -59,7 +59,7 @@ if ( ! function_exists( '_wp_can_use_pcre_u' ) ) {
 }
 
 // If _is_utf8_charset is already loaded.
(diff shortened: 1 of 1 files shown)


afragen:
Coding standards CI failure is fixed in other PRs.

afragen:
Unfortunately this was not just a simple drop-in. I needed to add more `! function_exists()` logic.

afragen:
Tag, you're it for that.

PR #165, opened 2025-07-16
Make `compat.php` compliant with the coding standard.

costdev:
`inc/compatibility/compat.php` is taken from WordPress Core. We'll probably be keeping this file somewhat in sync with WordPress Core's file in the future. However, it has some changes, such as additional `function_exists()` checks.

This PR:
- replaces a blanket PHPCS exclusion with targeted exclusions for:
  1. namespacing - These are polyfill functions that need to stay within the global namespace.
  2. assignment in `while` conditions - There's no need to change the code to facilitate this, which would require additional checks.
- Fixes indentation, array syntax, and replaces some instances of alternate syntax for control structures (`endif`, etc.) that was used inconsistently in the file.

### DIFF
--- a/inc/compatibility/compat.php
+++ b/inc/compatibility/compat.php
@@ -98,7 +98,7 @@ if ( ! function_exists( '_is_utf8_charset' ) ) {
 	}
 }
 
-if ( ! function_exists( 'mb_substr' ) ) :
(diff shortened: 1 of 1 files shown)


costdev:

afragen (approved):

PR #166, opened 2025-07-16
Replace `strpos()` with newer `str_*()` functions.

costdev:
This PR replaces uses of `strpos()` with the corresponding `str_*()` function where applicable.

Polyfills for these functions were introduced in #163.


costdev:

PR #167, opened 2025-07-16
Bump the minimum WordPress version to 6.4.

costdev:
This PR bumps the minimum WordPress version to 6.4.

The codebase has calls to functions that were introduced later than our current minimum, WordPress 5.4. For example, `wp_(get_)admin_notice()` was introduced in WordPress 6.4.

83.2% of WordPress websites checked by dotorg/stats are using WordPress 6.4 or greater. 65.9% are using WordPress 6.8 - the latest major at this time.

Here are some of the most popular plugins and their minimum WordPress version:
- Jetpack: 6.7
- WooCommerce: 6.7
- Yoast SEO: 6.6
- Elementor: 6.5
- BuddyPress: 6.1
- Advanced Custom Fields: 6.0
- bbPress: 6.0
- WPForms: 5.5

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -18,7 +18,7 @@
 	<config name="testVersion" value="7.4-" />
 
 	<!-- Set the minimum support WordPress version -->
-	<config name="minimum_supported_wp_version" value="5.4"/>
+	<config name="minimum_supported_wp_version" value="6.4"/>
 
 	<!-- Set the text domain -->
 	<config name="text_domain" value="fair"/>
--- a/plugin.php
+++ b/plugin.php
@@ -5,7 +5,7 @@
  * Version: 0.3.0
  * Author: FAIR Contributors
  * License: GPLv2
- * Requires at least: 5.4
+ * Requires at least: 6.4
  * Requires PHP: 7.4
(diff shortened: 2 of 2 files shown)


costdev:

costdev:
An alternative to this PR is to copy the definition of functions introduced in later WordPress versions into `inc/compatibility/compat.php`. Note, however, that this would require synchronizing "live" functions such as `wp_(get_)admin_notice()`, which are more prone to changes than the polyfills WordPress Core has in its `compat.php` file.

afragen:
Maybe we should add https://github.com/eduardovillao/wp-since to the CI workflow?

costdev:
Would certainly be useful. It would be good if it took a polyfills file into account so we wouldn't have to add `@wp-since ignore` comments beside each usage of any WP functions we polyfill. I can't see that functionality described in the repo though :thinking:

afragen:
I don't think it takes a polyfills file as an ignore. I'm pretty sure we'd have to individually ignore some of those functions.

costdev:
It's definitely something we should look into.

afragen:
Perhaps superseded by #170

PR #170, opened 2025-07-17
Split compat.php into php-polyfill.php and wp-polyfill.php

afragen:
This PR splits the `compat.php` into WP and PHP polyfills.

It assumes a minimum PHP version of 7.4.

It assumes a minimum WordPress version of 5.3 as this is the first WordPress version compatible with PHP 7.4.

I have not adjusted the GitHub Actions or the plugin Requires headers for these data.


afragen:
Coding standards error is not in any of these files and is fixed in #156

phpcs.xml.dist
```diff
@@ -55,7 +55,6 @@
 	</rule>
 
 	<rule ref="HM.Functions.NamespacedFunctions.MissingNamespace">
-		<exclude-pattern>inc/compatibility/compat\.php</exclude-pattern>
```
costdev, line 58: Add the two new files as exclusions for this sniff. They're flagged for this in the Coding Standards workflow errors.

inc/compatibility/wp-polyfill.php
```diff
@@ -0,0 +1,340 @@
...
+		}
+
+		if ( is_string( $args['type'] ) ) {
+			$type = trim( $args['type'] );
+
+			if ( str_contains( $type, ' ' ) ) {
+				_doing_it_wrong(
+					__FUNCTION__,
+					sprintf(
+					/* translators: %s: The "type" key. */
```
costdev: Indent issue.

inc/compatibility/wp-polyfill.php
```diff
@@ -0,0 +1,340 @@
...
+						__( 'The %s key must be a string without spaces.' ),
```
costdev:
```suggestion
						__( 'The %s key must be a string without spaces.' ), // phpcs:ignore WordPress.WP.I18n.MissingArgDomain -- This intentionally uses WordPress Core's translation string.
```

costdev (changes requested):

phpcs.xml.dist
```diff
@@ -55,7 +55,6 @@
 	</rule>
 
 	<rule ref="HM.Functions.NamespacedFunctions.MissingNamespace">
```
Ipstenu, line 57:
```suggestion
	<exclude-pattern>inc/compatibility/php-polyfill\.php</exclude-pattern>
	<exclude-pattern>inc/compatibility/wp-polyfill\.php</exclude-pattern>
```

Is that what you meant @costdev ?

Ipstenu (dismissed):

phpcs.xml.dist
afragen: Those files aren't flagged it's from #156

phpcs.xml.dist
afragen: Since we're maintaining the polyfills, I don't think we want to exclude them.

phpcs.xml.dist
costdev: Yep they are flagged: https://github.com/fairpm/fair-plugin/actions/runs/16355852050/job/46213913065#step:6:14

phpcs.xml.dist
costdev: We don't want to exclude them entirely no. Just from that particular sniff as we did with the previous `compat.php` file. I dropped a link in the other comment thread to the PHPCS errors for these new files

costdev: The `exclude-pattern` lines are correct in the suggestion, but the `rule` line should stay so they're only excluded from that rule.

phpcs.xml.dist
afragen: Should be fixed for our files.

phpcs.xml.dist
afragen: I get it now.

phpcs.xml.dist
afragen: Now fixed.

inc/compatibility/wp-polyfill.php
```diff
@@ -0,0 +1,340 @@
...
+
+		if ( is_string( $args['type'] ) ) {
+			$type = trim( $args['type'] );
+
+			if ( str_contains( $type, ' ' ) ) {
+				_doing_it_wrong(
+					__FUNCTION__,
+					sprintf(
+						/* translators: %s: The "type" key. */
+						__( 'The %s key must be a string without spaces.' ),
```
costdev: [The fix](https://github.com/fairpm/fair-plugin/commit/25804db5b4d8518e002f7fd2e27dca3071624c6a) for this was removed by mistake in [a later commit](https://github.com/fairpm/fair-plugin/commit/d9da5f47449067c1dabd357712002034b31a6c55)

afragen:
I'll create another PR to fix the coding standards issues from https://github.com/fairpm/fair-plugin/actions/runs/16355570099

phpcs.xml.dist
afragen: Fixed.

costdev (approved):

afragen:
@Ipstenu can you re-review please?

costdev:
@Ipstenu This seems ready to merge to me. We can't merge while your "Changes Requested" review is there without a follow-up approval. Can you take a look and see if you're happy to approve this one?

inc/compatibility/php-polyfill.php
```diff
@@ -0,0 +1,212 @@
...
+ */
+
+if ( ! function_exists( 'str_contains' ) ) {
+	/**
+	 * Polyfill for `str_contains()` function added in PHP 8.0.
+	 *
+	 * Performs a case-sensitive check indicating if needle is
+	 * contained in haystack.
+	 *
+	 * @since 5.9.0
```
jdevalk: This goes for all `@since` tags here, should we make specific that we're talking about _WordPress_ version numbers in this `@since`? Could become non-self explanatory in the future...

inc/compatibility/wp-polyfill.php
```diff
@@ -0,0 +1,340 @@
...
+			if ( ! $post_status ) {
+				return false;
+			}
+		}
+
+		if (
+		! is_object( $post_status ) ||
+		$post_status->internal ||
+		$post_status->protected
+		) {
```
jdevalk, line 92: The new lines above don't make it more readable in my eyes. I'd either make it one line or indent within the `if`, but that might be taste.

inc/compatibility/php-polyfill.php
costdev: Yeah I suppose if we added our own polyfills. Though we haven't been using `@since` for our own code, we might end up doing so.

It wouldn't hurt to be more specific here, and it's just a quick find/replace from `@since` to `@since WordPress`.

inc/compatibility/wp-polyfill.php
costdev: Looks like an indentation issue. The condition lines are usually one level deeper, including elsewhere in the file.

inc/compatibility/php-polyfill.php
afragen: Done

inc/compatibility/wp-polyfill.php
afragen: Also,  more readable if the boolean leads the additional conditions.

PR #171, opened 2025-07-17
Fix code standards issues

afragen:
A couple of coding standards issues were missed in #123

### DIFF
--- a/inc/settings/namespace.php
+++ b/inc/settings/namespace.php
@@ -9,8 +9,6 @@ namespace FAIR\Settings;
 
 use const FAIR\Avatars\AVATAR_SRC_SETTING_KEY;
 
-/* phpcs:disable Squiz.PHP.EmbeddedPhp.ContentBeforeOpen, Squiz.PHP.EmbeddedPhp.ContentAfterOpen, Squiz.PHP.EmbeddedPhp.ContentBeforeEnd */
-
 /**
  * Bootstrap.
  */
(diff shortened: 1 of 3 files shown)


afragen:
Now only coding standards issue are fixed in #156

inc/settings/namespace.php
```diff
@@ -8,6 +8,7 @@
 namespace FAIR\Settings;
 
 use const FAIR\Avatars\AVATAR_SRC_SETTING_KEY;
+
 /* phpcs:disable Squiz.PHP.EmbeddedPhp.ContentBeforeOpen, Squiz.PHP.EmbeddedPhp.ContentAfterOpen, Squiz.PHP.EmbeddedPhp.ContentBeforeEnd */
```
costdev: We could probably move these into different `<rule ref="..."> ... <exclude-pattern> </rule>` entries in phpcs.xml.dist so that whole-file exclusions for a rule can be found in that single ruleset file.

costdev (approved):
One possible alternative way to resolve the "empty line after USE statement", but not a blocker.

inc/settings/namespace.php
afragen: I didn't add those to begin with but if you want to move them as exclusions, I can.

afragen:
First lunch

inc/settings/namespace.php
costdev: It's not a blocker for this, so I can follow up later if we just want to get these errors resolved for now.

afragen: Done.

PR #172, opened 2025-07-18
use transient as wp_cache_* only lasts one page load

afragen:
Apparently `wp_cache_*` only last through one page load. I'm pretty sure we want to cache these `wp_remote_get()` calls between multiple page loads.

Currently the transient is set for `DID_CACHE_LIFETIME` which is 5 minutes.

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -13,7 +13,7 @@ use WP_Error;
 
 const SERVICE_ID = 'FairPackageManagementRepo';
 const CONTENT_TYPE = 'application/json+fair';
-const CACHE_LIFETIME = 5 * MINUTE_IN_SECONDS;
+const CACHE_LIFETIME = 12 * HOUR_IN_SECONDS;
 
 // phpcs:disable WordPress.NamingConventions.ValidVariableName
 
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -181,8 +181,10 @@ function fetch_metadata_doc( string $url ) {
 			'timeout' => 7,
 		] );
 		$code = wp_remote_retrieve_response_code( $response );
-		if ( is_wp_error( $response ) || $code !== 200 ) {
+		if ( is_wp_error( $response ) ) {
 			return $response;
+		} elseif ( $code !== 200 ) {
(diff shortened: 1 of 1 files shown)


inc/packages/namespace.php
```diff
@@ -13,7 +13,7 @@
 
 const SERVICE_ID = 'FairPackageManagementRepo';
 const CONTENT_TYPE = 'application/json+fair';
-const DID_CACHE_LIFETIME = 5 * MINUTE_IN_SECONDS;
+const CACHE_LIFETIME = 5 * MINUTE_IN_SECONDS;
```
jdevalk: Someone smarter than me probably thought about this 5 minute cache lifetime, but... Doesn't that seem short? If we can cache something for 5 minutes, we _could_ also cache for, an hour, two hours? 12 hours?

inc/packages/namespace.php
```diff
@@ -171,14 +171,19 @@ function install_plugin( string $id, ?string $version = null, $skin ) {
...
 function fetch_metadata_doc( string $url ) {
-	$response = wp_remote_get( $url, [
-		'headers' => [
-			'Accept' => sprintf( '%s;q=1.0, application/json;q=0.8', CONTENT_TYPE ),
-		],
-		'timeout' => 7,
-	] );
-	if ( is_wp_error( $response ) ) {
-		return $response;
+	$response = get_site_transient( md5( $url ) );
```
jdevalk: We're doing `md5( $url )` twice here, I'm wondering if we could do that only once, and then maybe prefix it too so it's a bit more recognizable in the transient cache?

inc/packages/namespace.php
afragen: Agreed. I just don't know how long to cache it. The data is coming from the PLC library and the dev's repository node. I'm assuming neither would change frequently.

afragen: 6 or 12 hours is probably sufficient, but I would like more input.

joedolson: This can be changed later if it's a problem, but I think we're better off trying something long, then shortening it. I think 12 hours would be fine.

afragen: Increased to 12 hours.

costdev (approved):

PR #173, opened 2025-07-18
Error checking to ensure data exists

afragen:
Prevents an issue if there is no banner data.

The things I come across when setting up another test package.

### DIFF
--- a/inc/updater/class-updater.php
+++ b/inc/updater/class-updater.php
@@ -324,7 +324,7 @@ class Updater {
 			'file'             => $filename,
 			'url'              => $this->metadata->url ?? $this->metadata->slug,
 			'sections'         => (array) $this->metadata->sections,
-			'icons'            => get_icons( $this->release->artifacts->icon ),
+			'icons'            => isset( $this->release->artifacts->icon ) ? get_icons( $this->release->artifacts->icon ) : [],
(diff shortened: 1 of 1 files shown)


inc/updater/class-updater.php
```diff
@@ -325,7 +325,7 @@ public function get_update_data() {
 			'url'              => $this->metadata->url ?? $this->metadata->slug,
 			'sections'         => (array) $this->metadata->sections,
 			'icons'            => get_icons( $this->release->artifacts->icon ),
```
costdev: Should we do the same with icon just in case?

inc/updater/class-updater.php
```diff
@@ -325,7 +325,7 @@ public function get_update_data() {
...
-			'banners'          => get_banners( $this->release->artifacts->banner ),
+			'banners'          => isset( $this->release->artifacts->banner ) ? get_banners( $this->release->artifacts->banner ) : [],
```
costdev, line 328: Is `$this->release->artifacts->banner` already an array, or does it need to be cast to one?

afragen: It is doesn't exist it's null.

inc/updater/class-updater.php
afragen: Probably

inc/updater/class-updater.php
afragen: Also updated for consistency.

costdev (approved):

PR #180, opened 2025-07-22
Simpify getting of ReleaseDocument during updating

afragen:
We only really need the DID to create the ReleaseDocument for the packages. Eliminates much complexity.

Tested and seems to work for plugin.php, core-update.php, and auto-updates.

### DIFF
--- a/inc/updater/namespace.php
+++ b/inc/updater/namespace.php
@@ -20,16 +20,15 @@ const RELEASE_PACKAGES_CACHE_KEY = 'fair-release-packages';
  */
 function bootstrap() {
 	add_action( 'init', __NAMESPACE__ . '\\run' );
-	add_action( 'get_fair_package_data', __NAMESPACE__ . '\\get_fair_release_data', 10, 1 );
 }
 
 /**
- * Get FAIR ReleaseDocument data.
+ * Add FAIR ReleaseDocument data to cache.
  *
  * @param string $did DID.
  * @return void
  */
-function get_fair_release_data( string $did ) : void {
+function add_package_to_release_cache( string $did ) : void {
 	if ( empty( $did ) ) {
 		return;
 	}
--- a/inc/packages/class-upgrader.php
+++ b/inc/packages/class-upgrader.php
@@ -7,6 +7,8 @@
 
 namespace FAIR\Packages;
 
+use function FAIR\Updater\add_package_to_release_cache;
+
 use WP_Error;
 use WP_Upgrader;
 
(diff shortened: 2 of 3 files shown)


inc/updater/namespace.php
```diff
@@ -22,63 +20,23 @@
...
- * @param string $filepath Absolute file path to package.
- * @param string $type plugin|theme.
- *
  * @return void
  */
-function get_fair_document_data( $did, $filepath, $type ) : void {
-	$packages = [];
+function get_fair_release_data( $did ) : void {
 	$releases = wp_cache_get( RELEASE_PACKAGES_CACHE_KEY );
 	$releases = $releases ? $releases : [];
```
costdev:
```suggestion
 	$releases = wp_cache_get( RELEASE_PACKAGES_CACHE_KEY ) ?: [];
```
While we're here

inc/updater/namespace.php
```diff
@@ -22,63 +20,23 @@
...
-	$file = $type === 'plugin' ? plugin_basename( $filepath ) : dirname( plugin_basename( $filepath ) );
 
-	// phpcs:disable HM.Security.NonceVerification.Recommended
-	// During auto-update.
-	if ( wp_doing_cron() || ( defined( 'WP_CLI' ) && WP_CLI ) ) {
+	if ( $did ) {
```
costdev: When will this not be true?

afragen: Probably never, but isn't it worth the check?

costdev: Yeah might aswell - though I'd say either wrap everything in it, including `wp_cache_get()`, or just make it a guard at the start of the function so no code runs if it's falsy

inc/updater/namespace.php
```diff
@@ -22,63 +20,22 @@
...
-		if ( str_contains( $file, $package ) ) {
-			$releases[ $did ] = get_latest_release_from_did( $did );
-			wp_cache_set( RELEASE_PACKAGES_CACHE_KEY, $releases );
-			break;
-		}
+function get_fair_release_data( $did ) : void {
+	if ( empty( $did ) ) {
+		return;
 	}
+ 	$releases = wp_cache_get( RELEASE_PACKAGES_CACHE_KEY ) ?: [];
```
costdev:
```suggestion
	$releases = wp_cache_get( RELEASE_PACKAGES_CACHE_KEY ) ?: [];
```

inc/updater/namespace.php
```diff
@@ -22,63 +20,22 @@
...
-	}
-	// phpcs:enable
-
-	foreach ( $packages as $package ) {
-		if ( str_contains( $file, $package ) ) {
-			$releases[ $did ] = get_latest_release_from_did( $did );
-			wp_cache_set( RELEASE_PACKAGES_CACHE_KEY, $releases );
-			break;
-		}
+function get_fair_release_data( $did ) : void {
```
costdev:
```suggestion
function get_fair_release_data( string $did ) : void {
```

inc/updater/namespace.php
afragen: I fixed.

PR #182, opened 2025-07-23
Use installation strings when installing

costdev:
This PR enables the installation strings used to display feedback during the installation process.

Additionally, the default Core `Plugin_Installer_Skin` or `Theme_Installer_Skin` is used when installing a plugin or theme respectively. This requires the porting of the `Plugin_Upgrader::plugin_info()` and `Theme_Upgrader::theme_info()` methods to the `Packages\Upgrader` class, as these are used to output action links after the installation process has completed.


afragen:
screenshot_686

PR #183, opened 2025-07-23
Explicitly associate the install note with the ID input field.

costdev:
The install note is important to knowing how to fill out the ID input field.

This PR converts the existing "fair-direct-install__note" HTML class to an ID, and uses the ID in the `aria-describedby` attribute of the ID input field.


joedolson (approved):
Looks good!

PR #184, opened 2025-07-23
Refactor get_update_data() to use DID, move package functions to FAIR\Packages

afragen:
Moves `get_update_data()` outside of class Updater and into inc/packages/namespace.php. Passes DID as a parameter to create the other data, MetadataDocument and ReleaseDocument. Only missing piece is file name which can be obtained from https://github.com/fairpm/mini-fair-repo/pull/22.

Also moves, `get_icons()` and `get_banners()` as they are more package related.

In this way the update data for either `plugins_api(`) or `themes_api(`) can be obtained only by knowing the DID. This data will be needed for creating Install modals as well as View details modals.


afragen:
It makes more sense for `get_icons(`), `get_banners()`, and `get_update_data()` to be in `FAIR\Packages`

inc/updater/class-updater.php
```diff
@@ -204,12 +201,12 @@ public function repo_api_details( $result, string $action, stdClass $response )
...
 
 		// Exit if not our repo.
-		$slug_arr = [ $this->metadata->slug, $this->metadata->slug . '-' . get_did_hash( $this->metadata->id ) ];
+		$slug_arr = [ $this->metadata->slug, $this->metadata->slug . '-' . Packages\get_did_hash( $this->metadata->id ) ];
 		if ( ! in_array( $response->slug, $slug_arr, true ) ) {
 			return $result;
 		}
 
-		return (object) $this->get_update_data();
+		return (object) Packages\get_update_data( $this->metadata->id );
```
costdev: If `$this->metadata->id` matches `$this->did`, should we just change all `$this->metadata->id` references to `$this->did` in this class?

costdev (commented):
This seems good so far, but I'd like to run some tests with it before we merge into #71.

inc/updater/class-updater.php
afragen: I'll check but I think you're correct.

afragen: Done.

Ipstenu (approved):

PR #185, opened 2025-07-26
AJAXify the "Install Now" button.

costdev:
This PR makes the "Install Now" button in the modal use AJAX for consistency with Core. It does so without replacing any of Core's JS, which should reduce the opportunity for future compatibility issues with Core.


afragen:
Tested.

screenshot_690

PR #187, opened 2025-07-26
Remove code not used with JS FAIR package installation

afragen:
Since installing a FAIR package really doesn't work without JS, since we have the Direct Install working correctly with Core AJAX, there is no need for certain code, like `class-upgrader.php`


costdev:
So far, you've removed exactly what I removed last night when testing and I encountered no errors. Approving the current removals. Will re-review if there are more removals.

PR #188, opened 2025-07-26
Return WP_Error for incorrect DID

afragen:
If incorrect DID is entered in Direct Install a new WP_Error is returned.

screenshot_692

### DIFF
--- a/inc/packages/did/class-plc.php
+++ b/inc/packages/did/class-plc.php
@@ -67,6 +67,9 @@ class PLC implements DID {
 		if ( json_last_error() !== JSON_ERROR_NONE ) {
 			return new WP_Error( 'fair.packages.did.json_error', __( 'Unable to parse DID document response.', 'fair' ) );
 		}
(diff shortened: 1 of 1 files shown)


inc/packages/did/class-plc.php
```diff
@@ -67,6 +67,12 @@ public function fetch_document() {
 		if ( json_last_error() !== JSON_ERROR_NONE ) {
 			return new WP_Error( 'fair.packages.did.json_error', __( 'Unable to parse DID document response.', 'fair' ) );
 		}
+		if ( 200 !== wp_remote_retrieve_response_code( $response ) && property_exists( $data, 'message' ) ) {
+			return new WP_Error( 'fair.packages.did.fetch.error', esc_html( $data->message ) );
+		}
+		if ( $data->id !== $this->id ) {
```
costdev:
```suggestion
		if ( empty( $data->id ) || $data->id !== $this->id ) {
```
Just to ensure the `id` property is actually present. The error message is fine as-is for this too.

joedolson (approved):

PR #197, opened 2025-07-28
Remove Packages, Updater from feature flag

afragen:
This removes Packages and Updater from the feature flag to allow default usage of Install and Update of FAIR packages.

We still need a theme package to test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ The FAIR plugin implements federated or local versions of the following features
 * Browser and server health checks
 * Other APIs such as the Credits API, Secret Keys API, and Importers API
 * Twemoji images for emojis
-* Direct installation and updating of FAIR packages using DID
+* Installation and updating of packages direct from their source repository, without talking to any centralized server
 
 The default FAIR provider in this plugin is [AspireCloud from AspirePress](https://aspirepress.org/). The AspirePress team were key in helping the FAIR project get off the ground. As the FAIR project grows and other providers come online you will be able to configure your chosen FAIR provider within the plugin.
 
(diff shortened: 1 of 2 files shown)


inc/namespace.php
```diff
@@ -33,17 +33,18 @@ function bootstrap() {
...
 	Salts\bootstrap();
 	Settings\bootstrap();
+	Updater\bootstrap();
 	Upgrades\bootstrap();
 	User_Notification\bootstrap();
 	Version_Check\bootstrap();
 
+	if ( defined( 'FAIR_EXPERIMENTAL_PACKAGES' ) && FAIR_EXPERIMENTAL_PACKAGES ) {
+	}
+
```
rmccue: This should be removed by the looks of it?

README.md
```diff
@@ -35,6 +35,7 @@ The FAIR plugin implements federated or local versions of the following features
 * Browser and server health checks
 * Other APIs such as the Credits API, Secret Keys API, and Importers API
 * Twemoji images for emojis
+* Direct installation and updating of FAIR packages using DID
```
rmccue: I would move this to the top of the list and rephrase to
```suggestion
* Installation and updating of packages direct from their source repository, without talking to any centralized server
```

README.md
```diff
@@ -49,7 +50,7 @@ In addition to the key FAIR implementations, a few other features in WordPress a
 
 As FAIR works towards our plans for full decentralized package management, some features are marked as experimental. These features must be manually opted-in to during development.
 
-* `FAIR_EXPERIMENTAL_PACKAGES` - Define as `true` to enable decentralized package installation, via direct DID input.
+* `FAIR_EXPERIMENTAL_PACKAGES` - Define as `true` to enable any experimental features.
```
rmccue: I'd remove this whole section now that we're not using it. (We'd use a different name that didn't have `_PACKAGES` in the name for other features.)

PR #198, opened 2025-07-29
Fix context message for platform string.

joedolson:
This is a re-build of @marcarmengou's PR.


marcarmengou:
Thanks for the re-build @joedolson!

PR #218, opened 2025-08-14
Use wp_cache_* instead of get|set_transient

afragen:
For consistency let's use wp_cache_* instead of transients.

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -15,6 +15,7 @@ use WP_Upgrader;
 
 const SERVICE_ID = 'FairPackageManagementRepo';
 const CONTENT_TYPE = 'application/json+fair';
+const CACHE_KEY = 'fair-packages-';
 const CACHE_LIFETIME = 12 * HOUR_IN_SECONDS;
 const RELEASE_PACKAGES_CACHE_KEY = 'fair-release-packages';
 
@@ -83,7 +84,7 @@ function get_did_hash( string $id ) {
  * @return DIDDocument|WP_Error
  */
 function get_did_document( string $id ) {
-	$cached = get_site_transient( $id );
+	$cached = wp_cache_get( $id );
 	if ( $cached ) {
 		return $cached;
 	}
@@ -98,7 +99,7 @@ function get_did_document( string $id ) {
 	if ( is_wp_error( $document ) ) {
 		return $document;
 	}
-	set_site_transient( $id, $document, CACHE_LIFETIME );
+	wp_cache_set( $id, $document, 'did', CACHE_LIFETIME );
 
 	return $document;
 }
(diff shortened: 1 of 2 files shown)


costdev:
Since not all sites will have a persistent object cache, using `wp_cache_*()` isn't as suitable as the Transient API for things we _definitely_ want to be cached beyond a single page load.

For example, `php_releases` seems better in the Transient API, since the chance of that updating before the cache expires is near zero. I think the other changes to use `wp_cache_*()` are fine though, since we probably do want those to get the latest available data from a DID document / metadata document when a persistent object cache isn't used.

afragen:
Does anyone not have a persistent object cache these days?

costdev:
I'd say quite a lot yeah.

inc/packages/namespace.php
```diff
@@ -83,7 +84,7 @@ function get_did_hash( string $id ) {
  * @return DIDDocument|WP_Error
  */
 function get_did_document( string $id ) {
-	$cached = get_site_transient( $id );
+	$cached = wp_cache_get( $id );
```
```suggestion
	$cached = wp_cache_get( $id, 'did' );
```
Also, should this be `did-docs` for consistency with `metadata-docs`?

inc/packages/namespace.php
```diff
@@ -142,8 +143,8 @@ function fetch_package_metadata( string $id ) {
  * @return MetadataDocument|WP_Error
  */
 function fetch_metadata_doc( string $url ) {
-	$cache_key = md5( $url );
-	$response = get_site_transient( $cache_key );
+	$cache_key = CACHE_KEY . md5( $url );
+	$response = wp_cache_get( $cache_key );
```
```suggestion
	$response = wp_cache_get( $cache_key, 'metadata-docs' );
```

inc/packages/namespace.php
afragen: Does the group need to be present on `wp_cache_get()`?

afragen: I guess it does if it's declared.

afragen: Or should we just remove it the group?

inc/packages/namespace.php
```diff
@@ -98,7 +99,7 @@ function get_did_document( string $id ) {
 	if ( is_wp_error( $document ) ) {
 		return $document;
 	}
-	set_site_transient( $id, $document, CACHE_LIFETIME );
+	wp_cache_set( $id, $document, 'did', CACHE_LIFETIME );
```
costdev: Not for this PR, but I'm wondering if we should actually be caching the HTTP response in the `::fetch_document()` method rather than the document object here. That would mean anything calling `::fetch_document()` could benefit from the cache, and would make it consistent with the metadata cache, which caches the HTTP response, not the document object.

inc/packages/namespace.php
costdev: Since this just uses `$id` as the cache key, it's probably best to keep the groups.

inc/packages/namespace.php
afragen: I could write a function similar to the one in Mini-FAIR for all the `wp_remote_get` calls. Cache every remote call.

Ipstenu (approved):
Looks like you made all the changes!

kasparsd:
I wanted to `+1` what @costdev mentioned before -- not all hosts have persistent object cache available. Transients actually use persistent object cache if available but not the other way around. I would recommend reverting back to transients for this.

On sites with many plugins and themes the lack of caching will add a significant delay to the update/install screens as many of the checks happen synchronously on those screens.

afragen:
https://github.com/fairpm/mini-fair-repo/pull/46
https://github.com/fairpm/fair-plugin/pull/248

Issue #219, opened 2025-08-14
Issue with FAIR and Mini-FAIR


afragen:
There seems to be an issue with FAIR and Mini-FAIR being active at the same time when the site running Mini-FAIR has registered packages. Something seems to grind to a halt and I'm not quite sure what or where that is at the moment.

All I can say for sure is, at the moment, don't run FAIR on a site you're using to register packages.

I'm trying to get some things setup for testing locally.

afragen:
Here's the root of the problem. `wp_remote_get()` often returns a timeout error when trying to get the Mini-FAIR endpoint from the same site.

<img width="1310" height="59" alt="Image" src="https://github.com/user-attachments/assets/e2184ce3-b0c5-4800-af93-2bbd1a35ce5c" />

afragen:
@rmccue given your insight into `Requests` perhaps you can discern why this happens.

I was able to create a workaround. It's not pretty but it works.

There are 2 parts.

1. From the Mini-FAIR endpoint I cache the result.
2. From `fetch_metadata_doc()`, if the `home_URL()` is in the endpoint URL, I create a new `$response` with just enough data to cache and send to `MetadataDocument::from_response( $response )`

```php
	// Huge kludge.
	if ( ! $response ) {
		if ( str_contains( $url, home_url() ) ) {
			$did = explode( '/', parse_url( $url, PHP_URL_PATH ) );
			$did = array_pop( $did );
			$body = wp_cache_get( 'rest-endpoint-' . $did );
			$response  = [];
			$response['headers'] = [];
			$response['body'] = json_encode( $body );
			$response = ! $body ? false : $response;
			if ( $response ) {
				wp_cache_set( $cache_key, $response, 'metadata-docs', CACHE_LIFETIME );
			}
		}
	}
```

afragen:
I coded this in a function. It's sitting in https://github.com/afragen/fair-plugin/tree/local-metadata

~~It needs #218 and https://github.com/fairpm/mini-fair-repo/pull/38.~~ It also needs the site running Mini-FAIR to use an persistent object cache.

rmccue:
I've seen this happen before; it's a broader loopback connection issue. Does Site Health highlight that it can't create loopback requests properly? (Your wp-cron would also break if so.)

afragen:
Loopback requests are functioning and my wp-cron works.

afragen:
Closed with #221

PR #221, opened 2025-08-15
Use local metadata when attempting to get Mini-FAIR data from same site as package is registered

afragen:
Re: #219 

When using FAIR and Mini-FAIR on same site, there is a timeout issue when attempting to to retrieve the Mini-FAIR REST endpoint. 

What I've done is cached the Mini-FAIR Metadata endpoint ( https://github.com/fairpm/mini-fair-repo/pull/38/commits/e899132ba6997f53d349357b8109c0b7f465e596 ) in https://github.com/fairpm/mini-fair-repo/pull/38 and used this cached data to add the data into `fetch_metadata_doc()`

This PR also contains #218. I would recommend merging #218 first.

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -148,11 +148,17 @@ function fetch_metadata_doc( string $url ) {
 	$response = fetch_metadata_from_local( $response, $url );
 
 	if ( ! $response ) {
-		$response = wp_remote_get( $url, [
+		$options = [
 			'headers' => [
 				'Accept' => sprintf( '%s;q=1.0, application/json;q=0.8', CONTENT_TYPE ),
 			],
-		] );
+		];
+
+		// Set low timeout for local package.
+		if ( str_contains( $url, home_url() ) ) {
+			$options['timeout'] = 1;
+		}
+		$response = wp_remote_get( $url, $options );
 		$code = wp_remote_retrieve_response_code( $response );
 		if ( is_wp_error( $response ) ) {
 			return $response;


inc/packages/namespace.php
```diff
@@ -145,6 +145,7 @@ function fetch_package_metadata( string $id ) {
 function fetch_metadata_doc( string $url ) {
 	$cache_key = CACHE_KEY . md5( $url );
 	$response = wp_cache_get( $cache_key, 'metadata-docs' );
+	$response = fetch_metadata_from_local( $response, $url );
```
costdev, line 148: We can refactor the `if` block to only make calls when necessary and make things a bit cleaner:

```php
if ( ! $response ) {
	if ( str_contains( $url, home_url() ) ) {
		$response = fetch_metadata_from_local( $url );
	} else {
		$response = wp_remote_get( $url, [
			'headers' => [
				'Accept' => sprintf( '%s;q=1.0, application/json;q=0.8', CONTENT_TYPE ),
			],
			'timeout' => 7,
		] );

		if ( wp_remote_retrieve_response_code( $response ) !== 200 ) {
			$response = new WP_Error( 'fair.packages.metadata.failure', __( 'HTTP error code received', 'fair' ) );
		}
	}

	if ( is_wp_error( $response ) ) {
		return $response;
	} 

	wp_cache_set( $cache_key, $response, 'metadata-docs', CACHE_LIFETIME );
}
```
- Saves on needing the `$response` parameter in `fetch_metadata_from_local()`.
- Saves on the `! $response` check and the `str_contains()` check in `fetch_metadata_from_local()`.
- Ensures `$response = false` is never passed to MetadataDocument::from_response()`.

- `fetch_metadata_from_local()` should become `@return array|WP_Error`, and return an appropriate error should `wp_cache_get( 'fair-metadata-endpoint-' . $did, 'metadata-endpoints' )` return `false`.

inc/packages/namespace.php
```diff
@@ -165,6 +166,35 @@ function fetch_metadata_doc( string $url ) {
...
+			$response['body'] = json_encode( $body );
+			$response = ! $body ? false : $response;
+			if ( $response ) {
+				wp_cache_set( CACHE_KEY . md5( $url ), $response, 'metadata-docs', CACHE_LIFETIME );
+			}
+		}
+	}
+
+	return $response;
+}
```
costdev, line 196: With the above suggestion, `fetch_metadata_from_local()` could then become this:

```suggestion
 * @return array|WP_Error
 */
function fetch_metadata_from_local( $url ) {
	$did = explode( '/', parse_url( $url, PHP_URL_PATH ) );
	$did = array_pop( $did );
	
	$body = wp_cache_get( 'fair-metadata-endpoint-' . $did, 'metadata-endpoints' );
	if ( ! $body ) {
		return new WP_Error( 'fair.packages.metadata.failure', __( 'Local metadata not available.', 'fair' ) );
	}

	$response = [
		'headers' => [],
		'body' => json_encode( $body ),
	];
	wp_cache_set( CACHE_KEY . md5( $url ), $response, 'metadata-docs', CACHE_LIFETIME );
	return $response;
}
```
- Removed `$response` parameter.
- Removed `! $response` and `str_contains()` checks.
- Bailed with `WP_Error` when there's no metadata cached.
- Merged the `$response` array's generation into one block.

afragen:
For the record, @costdev and I did try to directly use the REST endpoint to return the values, but for some reason some of the data was missing. Not sure why, but it did make it not a viable option.

costdev:
> For the record, @costdev and I did try to directly use the REST endpoint to return the values, but for some reason some of the data was missing. Not sure why, but it did make it not a viable option.

Following up: We looked at what would be needed to get the remaining data. It would require a couple of extra calls in the Mini FAIR repo plugin. It's possible that this could cause delays outside of the circumstances described in this PR. While we haven't verified this, there was a mutual feeling that trying to pursue a direct call was complicating matters further.

@afragen looked at reducing the `timeout` argument of the `wp_remote_get()` call when it's to the same site, and that appears to be testing well.

PR #223, opened 2025-08-16
Fix dashboard news widget and add caching

### LINKED ISSUE #222: Dashboard News widget doesn't seem to be loading properly
While looking to cache all the `wp_remote_get` functions I could, I came across the `inc/dashboard-widgets/namespace.php` file.

I discovered an early exit that was supposed to allow for compatibility with ClassicPress with the following code in `bootstrap()`.

```php
// Support existing software like ClassicPress, which removes this feature.
	if ( ! function_exists( 'wp_print_community_events_markup' ) ) {
		return;
	}
```

What I discovered was that the function doesn't seem to be defined at the point that it is called and so the entire `bootstrap()` code is skipped.

I was able to move the guard closer to where it is actually needed and I think this provides ClassicPress compatibility while allowing the code to run.

When events are picked up from [thewp.world](https://thewp.world/) there is a notification at the bottom of the widget.

Image

afragen:
The `function_exists` check needs to be moved as function not yet defined and so the entire `inc\dashboard-widgets/namespace.php::bootstrap()` is skipped.

This also adds caching to the `wp_remote_get()` call.

Fixes #222

### DIFF
--- a/inc/dashboard-widgets/namespace.php
+++ b/inc/dashboard-widgets/namespace.php
@@ -84,13 +84,13 @@ function get_community_events_ajax() : void {
  * @return array|WP_Error List of events or WP_Error on failure.
  */
 function get_community_events() {
-	$response = wp_cache_get( EVENTS_API );
+	$response = get_transient( EVENTS_API );
 	if ( ! $response ) {
 		$response = wp_remote_get( EVENTS_API );
 		if ( is_wp_error( $response ) ) {
 			return $response;
 		}
-		wp_cache_set( EVENTS_API, $response, '', CACHE_LIFETIME );
+		set_transient( EVENTS_API, $response, CACHE_LIFETIME );
 	}
 
 	$data = json_decode( wp_remote_retrieve_body( $response ), true );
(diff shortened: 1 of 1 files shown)


inc/dashboard-widgets/namespace.php
```diff
@@ -86,9 +84,13 @@ function get_community_events_ajax() : void {
  * @return array|WP_Error List of events or WP_Error on failure.
  */
 function get_community_events() {
-	$response = wp_remote_get( EVENTS_API );
-	if ( is_wp_error( $response ) ) {
-		return $response;
+	$response = wp_cache_get( EVENTS_API );
```
costdev: We should use the Transient API here instead. Sites without a persistent cache won't see a benefit when this isn't likely to get hit many times in one page load.

costdev (changes requested):
Just one change request, otherwise looks good

inc/dashboard-widgets/namespace.php
afragen: Actually, even without a persistent cache `WP_Cache_Object` lasts for a page load. It's subsequent page loads that are affected.

costdev: That's what I mean - the non-persistent cache only lasts for a page load, so using wp_cache_* wouldn't give any benefit for those sites, so the Transient API is better here

costdev (approved):

PR #234, opened 2025-08-26
Supporting mu-plugins usage > Option 1: Update PLUGIN_DIR definition for mu-plugins context awareness

sc0ttkclark:
# Supporting mu-plugins usage

* Option 1: [Supporting using plugin.php as the mu-plugin loader](https://github.com/fairpm/fair-plugin/pull/234) (this PR)
* Option 2: [Add an mu-plugin stub file with the same plugin headers](https://github.com/fairpm/fair-plugin/pull/235)

## Summary

I would like hosting platforms to be more easily able to copy the `plugin.php` into the `mu-plugins` directory and have it reference the subdirectory `fair-plugin` more easily.

## Benefits

* FAIR plugin can be loaded as an mu-plugin
* It appears as Must-Use automatically in the plugins area
* The current plugin header information can be utilized in the plugins area which includes version number
* Ease of turning the plugin into an mu-plugin
* No custom [mu-plugins loader](https://github.com/jazzsequence/plague-music-wp-upstream/blob/main/wp-content/mu-plugins/loader.php) or [custom plugin loader (as MU)](https://gist.github.com/afragen/9117fd930d9be16be8a5f450b809dfa8) needed

## Cons

* FAIR plugin code itself is modified to change which directory it uses based on whether in mu-plugins context
* FAIR plugin code hardcodes the fair-plugin slug when used in mu-plugins context

### Example in "Must-Use" section

FAIR plugin in the Must-Use plugins section

### Example in "Must-Use" section when just a PHP file including the plugin

FAIR plugin being included by an mu-plugins loader, shown in the Must-Use plugins section

### DIFF
--- a/plugin.php
+++ b/plugin.php
@@ -32,29 +32,29 @@ if ( __DIR__ === WPMU_PLUGIN_DIR ) {
 define( __NAMESPACE__ . '\PLUGIN_DIR', $fair_dir );
 
 // Include core.
-require_once __DIR__ . '/inc/namespace.php';
-require_once __DIR__ . '/inc/avatars/namespace.php';
-require_once __DIR__ . '/inc/credits/namespace.php';
-require_once __DIR__ . '/inc/dashboard-widgets/namespace.php';
-require_once __DIR__ . '/inc/default-repo/namespace.php';
-require_once __DIR__ . '/inc/disable-openverse/namespace.php';
-require_once __DIR__ . '/inc/icons/namespace.php';
-require_once __DIR__ . '/inc/importers/namespace.php';
-require_once __DIR__ . '/inc/packages/namespace.php';
-require_once __DIR__ . '/inc/packages/admin/namespace.php';
-require_once __DIR__ . '/inc/packages/admin/info.php';
-require_once __DIR__ . '/inc/pings/namespace.php';
-require_once __DIR__ . '/inc/salts/namespace.php';
-require_once __DIR__ . '/inc/settings/namespace.php';
-require_once __DIR__ . '/inc/upgrades/namespace.php';
-require_once __DIR__ . '/inc/updater/namespace.php';
-require_once __DIR__ . '/inc/user-notification/namespace.php';
-require_once __DIR__ . '/inc/version-check/namespace.php';
+require_once PLUGIN_DIR . '/inc/namespace.php';
+require_once PLUGIN_DIR . '/inc/avatars/namespace.php';
+require_once PLUGIN_DIR . '/inc/credits/namespace.php';
+require_once PLUGIN_DIR . '/inc/dashboard-widgets/namespace.php';
+require_once PLUGIN_DIR . '/inc/default-repo/namespace.php';
+require_once PLUGIN_DIR . '/inc/disable-openverse/namespace.php';
+require_once PLUGIN_DIR . '/inc/icons/namespace.php';
+require_once PLUGIN_DIR . '/inc/importers/namespace.php';
+require_once PLUGIN_DIR . '/inc/packages/namespace.php';
+require_once PLUGIN_DIR . '/inc/packages/admin/namespace.php';
+require_once PLUGIN_DIR . '/inc/packages/admin/info.php';
+require_once PLUGIN_DIR . '/inc/pings/namespace.php';
+require_once PLUGIN_DIR . '/inc/salts/namespace.php';
+require_once PLUGIN_DIR . '/inc/settings/namespace.php';
+require_once PLUGIN_DIR . '/inc/upgrades/namespace.php';
+require_once PLUGIN_DIR . '/inc/updater/namespace.php';
+require_once PLUGIN_DIR . '/inc/user-notification/namespace.php';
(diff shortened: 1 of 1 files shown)


sc0ttkclark:
Confirmed that this is working on the managed hosting platform I'm testing it for. I installed `did:plc:afjf7gsjzsqmgc7dlhb553mv` for gitupdater and it worked like a treat!

plugin.php
```diff
@@ -19,32 +19,33 @@
 namespace FAIR;
 
 const VERSION = '0.4.1';
-const PLUGIN_DIR = __DIR__;
+const PLUGIN_DIR = __DIR__ === WPMU_PLUGIN_DIR ? __DIR__ . '/fair-plugin' : __DIR__;
```
kasparsd, line 22: This is making an assumption that the plugin directory is named `fair-plugin`. When installed via DID, the plugin folder would actually be dynamic and match the DID.

sc0ttkclark:
> Is the goal here to have the correct meta information displayed in the wp-admin when viewing the must-use plugin list for the “loader” plugin while keeping the plugin under wp-content/plugins/fair-plugin?

Yes exactly

> If so, could we instead ship a loader stub as part of the fair plugin which can be required from the mu-plugins/your-loader.php, and would inject the correct meta data via the filters instead?

I would love that too, just unsure the logistics of keeping headers in sync within the fair-plugin release process.

> Coupling the plugin logic to assumptions about global environment and path relationships has proven to be fragile in my experience with enterprise setups.

Agree, this is just one PR option and I will work on another PR for the stub option.

plugin.php
sc0ttkclark: Only in the mu-plugins context, unsure if FAIR plugin or WordPress itself would ever support installing plugins into mu-plugins though.

kasparsd: From the change here I read that the manual install method to `wp-content/mu-plugins` involves:

1. Using Composer or some other mechanism to place the `fair-plugin` directory at `wp-content/mu-plugins/fair-plugin`.

2. Copying/moving the `plugin.php` from plugin directory to `wp-content/mu-plugins/plugin.php` (most likely renamed to something like `fair-plugin.php`).

Thats a lot of moving parts and assumptions. For example, the plugin directory name under `mu-plugins` can only be `fair-plugin`. Placing it under `wp-content/mu-plugins/fair` would break the bootstrap logic.

What happens with the bundled Git updater logic when this setup is used? By default, WP core doesn't handle automatic updates for mu-plugins. I haven't checked that.

sc0ttkclark: Updates would be managed on the platform or if done via composer itself, when relying on the mu-plugins usage

sc0ttkclark: The plugin slug I went with mirrors that of the GitHub repo slug itself which is a _fair_ assumption (see what I did there? 😂)

kasparsd: I wonder if this would benefit from a GitHub issue to collect and define all the use-cases we wan't to support, and how they impact the core functionality of this plugin. The issue with hard-coded `__DIR__ . '/fair-plugin'` would still apply to anyone doing it differently.

If there is a process being implemented for updating this under `mu-plugins`, isn't a simplest option to introduce a loader file that (1) bootstraps the fair plugin, and (2) hides the activation controls in the regular plugin view?

This is how VIP does it for all plugins activated via their [`wpcom_vip_load_plugin()` helper](https://github.com/Automattic/vip-go-mu-plugins/blob/82b0b9b300e3654920729b125b11dd1e76cd0e21/vip-helpers/vip-utils.php#L1093-L1252), for example.

sc0ttkclark: You would think an mu plugins loader would be built into WordPress core already and you just have a file you include to run the logic. Even .com has this code already but alas..

I'm open to other options too, whatever we can do to lower the barrier for more hosting platforms to include FAIR.

kasparsd: Could you describe the current logic of that loader file that you're already using? I've done a bunch for Pantheon and VIP but they are mostly just `require_once dirname( __DIR__ ) . '/plugins/plugins-slug/plugin.php` since it happens in a single loader file and I don't need to clone the meta headers. 

Here is [how Pantheon loads their own must-use plugin](https://github.com/pantheon-systems/WordPress/blob/default/wp-content/mu-plugins/loader.php) which is added via Composer (as `wordpress-mu-plugin` type). I guess that is what your other example was from.

I feel like every plugin would suffer from this issue so it feels like best solved at the host level where they introduce a standardized way for all their plugins which are added as mu-plugins.

rmccue: Huh, neat, Pantheon copied my code without crediting it :D The original loader for that is here: https://github.com/humanmade/hm-base/blob/master/content/mu-plugins/loader.php

I agree that this is likely best handled by the platform, but I think we can also help assist with that where it makes sense, provided it doesn't increase the burden to us.

sc0ttkclark:
Looks like there's no appetite by others to build in direct support for using `fair-plugin` as an mu-plugin with copying a file. I'll work on supporting this at the platform level instead.

PR #235, opened 2025-08-26
Supporting mu-plugins usage > Option 2: Add an mu-plugin stub file with the same plugin headers

sc0ttkclark:
# Supporting mu-plugins usage

* Option 1: [Supporting using plugin.php as the mu-plugin loader](https://github.com/fairpm/fair-plugin/pull/234)
* Option 2: [Add an mu-plugin stub file with the same plugin headers](https://github.com/fairpm/fair-plugin/pull/235) (this PR)

## Summary

This is the flip side of https://github.com/fairpm/fair-plugin/pull/234 which instead of updating the plugin file to support being used as the mu-plugin file itself, this will be a separate stub file that could then be copied/used by hosting environments as the mu-plugin loader and includes plugin header information.

I would like hosting platforms to be more easily able to copy the mu-plugin stub into the `mu-plugins` directory and have it reference the subdirectory `fair-plugin` more easily.

## Benefits

* FAIR plugin can be loaded as an mu-plugin
* It appears as Must-Use automatically in the plugins area
* The current plugin header information can be utilized in the plugins area which includes version number
* Ease of turning the plugin into an mu-plugin
* No custom [mu-plugins loader](https://github.com/jazzsequence/plague-music-wp-upstream/blob/main/wp-content/mu-plugins/loader.php) or [custom plugin loader (as MU)](https://gist.github.com/afragen/9117fd930d9be16be8a5f450b809dfa8) needed

## Cons

* Stub file headers have to stay in sync with `plugin.php`
* Stub file hardcodes the fair-plugin slug

### Example in "Must-Use" section

FAIR plugin in the Must-Use plugins section

### Example in "Must-Use" section when just a PHP file including the plugin

FAIR plugin being included by an mu-plugins loader, shown in the Must-Use plugins section

### DIFF
--- a/inc/stub-mu-plugin.php
+++ b/inc/stub-mu-plugin.php
@@ -16,4 +16,9 @@
  * Network: true
  */
 
+/**
+ * This is a stub file you can use for your mu-plugins implementation of the FAIR plugin.
+ *
+ * Copy this into wp-content/mu-plugins/fair-plugin.php and modify as needed.
+ */
 require_once __DIR__ . '/fair-plugin/plugin.php';


sc0ttkclark:
Looks like there's no appetite by others to build in direct support for using `fair-plugin` as an mu-plugin with copying a file. I'll work on supporting this at the platform level instead.

Issue #236, opened 2025-08-26
Register fair-plugin with DID


afragen:
At some point we will need to register the `fair-plugin` with a DID and then it can update without a dependency on Git Updater Lite.

afragen:
This really only needs the server at `api.fair.pm` to be running the Mini-FAIR plugin so we can register the FAIR plugin.

cdils:
Agreed on today's call to try the GitHub Action for this

namithj:
The GH Action is added in here https://github.com/fairpm/fair-plugin/pull/422

How do we go about with the actual release, run the action on 1.2.2 or do a 1.2.3 as per schedule?
We need to set the releases as mutable on the repo so that the action can write the metadata.json file

namithj:
GH Action Workflow PR is withdrawn pending consensus on which method to use.

afragen:
Awaiting update of Beacon to use DID Manager and possible rearchitect of plugin.

kasparsd:
I believe the DID was added to the plugin in #482, pushed to https://web.plc.directory/did/did:plc:ppicmk23c5pimdivve34bcp2 and the metadata is being published via GitHub action added in a88aefa58f49dd2f6c8ea9d3b7439695a91cb181 to https://github.com/fairpm/fair-plugin/releases/latest/download/fair-metadata.json

Is there anything else left to be done?

PR #237, opened 2025-08-26
Remove Git Updater Lite dependency

afragen:
Do NOT merge until `fair-plugin` has a registered DID and it is set in the plugin headers.

This will remove the GIt Updater Lite dependency for updating the `fair-plugin`.

See fairpm/fair-plugin#236


costdev:
Remove the `require_once` from `plugin.php` too.

costdev (dismissed):

costdev:
# For other reviewers

As stated in the PR's description:
> Do NOT merge until `fair-plugin` has a registered DID and it is set in the plugin headers.

Ipstenu:
@afragen I'm flagging this for 1.2 :)

costdev:
Blocked by fairpm/fair-plugin#236

kasparsd:
I believe since the 1.4 the plugin is using the DID update mechanism instead of the bundled `( new Git_Updater\Lite( PLUGIN_FILE ) )->run()` since the DID updater is hooked _after_ the lite into `site_transient_update_plugins` and takes over the response.

Secondly, the git-updater doesn't appear to work with the DID suffixed plugin directories like https://api.fair.pm/git-updater/v1/update-api/?slug=fair-32405b (returns `Specified repo does not exist`) so it would fail for all setups where the plugin directly was recently moved (on 1.4 upgrade).

afragen:
@kasparsd Git Updater does work with hashed slugs in plugins or themes directories. 

We were not yet removing this as there was concern that if we did, anyone still on a version earlier than FAIR Connect 1.4.1 wouldn't see any updates.

afragen:
That endpoint is accurate as there is not an installed repository at `wp-content/plugins/fair-32405b` in that site.

kasparsd:
> Git Updater does work with hashed slugs in plugins or themes directories.

@afragen Can you please elaborate on this? On future upgrades from v1.4 the DID update path is used which also triggers the directory rename to include the did-hash. I confirmed this by lowering a v1.4 install to `v1.3.9` which then upgraded to v1.4 using the v1.4 code, and it used the DID upgrade path (and applied the plugin directory rename).

after-1 4-upgrades

---

 The `Git_Updater\Lite` in v1.4+ still triggers but it now uses the `fair-32405b` slug against the API https://api.fair.pm/wp-json/git-updater/v1/update-api/?slug=fair-32405b which returns:

```json
{"error":"Specified repo does not exist."}
```

Even if it did return a response, the lifecycle order of hooks makes the DID response override the `git-updater` response effectively making it not functional.

So older versions with bundled `Git_Updater\Lite` can still function and work but for anything above v1.4 it isn't really needed or used.

> anyone still on a version earlier than FAIR Connect 1.4.1 wouldn't see any updates.

Can you please elaborate on that? Every version of FAIR connect until (and including) 1.4 has the `Git_Updater\Lite` and can update using the bundled `Git_Updater\Lite` in those versions. We definitely need to keep the `Git_Updater\Lite` API endpoints around but the library isn't really needed in the plugin.

afragen:
I thought we fixed it so that if a package was installed via DID it would retain the hashed slug and if it was installed by any other means then the slug would stay unhashed. This would prevent any updates from resulting in a deactivated plugin. This was previously tested and working. 

Can you provide explicit instructions on how your testing changed the slug? Was it due to new code changes?

kasparsd:
@afragen I just reported it here fairpm/fair-plugin#502. The instructions are in there -- when on 1.4.0 (installed via ZIP) I fake-downgrade to 1.3.9 and run the update.

afragen:
I'm confused. It appears that your installation of the plugin is incorrect. 

The plugin slug should be `fair-plugin` or `fair-plugin-didhash`. What you have is neither of those.

PR #243, opened 2025-09-01
Add package domain validation

### LINKED ISSUE #242: Establish Validation via DNS
Include DNS verification - reporting domain is domain

Note: Verification isn't the right word.

rmccue:
Fixes #242.

Subject to review from trust perspective, and the microcopy here needs some iteration.

Valid | Invalid | Missing
-- | -- | --
Screenshot 2025-09-01 at 21 42 43 | Screenshot 2025-09-01 at 21 39 53 | Screenshot 2025-09-01 at 21 29 29

### DIFF
--- a/inc/packages/admin/info.php
+++ b/inc/packages/admin/info.php
@@ -448,7 +448,7 @@ function add_requirement_notices( ReleaseDocument $release ) : void {
  */
 function render_alias_notice( DIDDocument $did ) : bool {
 	$validation = Packages\validate_package_alias( $did );
-	$title = __( 'Domain Alias', 'fair' );
+	$title = __( 'Domain Alias:', 'fair' );
 	$result = false;
 	switch ( gettype( $validation ) ) {
 		case 'string':
@@ -497,7 +497,7 @@ function render_alias_notice( DIDDocument $did ) : bool {
 	}
 
 	printf(
-		'<strong>%s</strong>: %s',
+		'<strong>%s</strong> %s',
 		esc_html( $title ),
 		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- Sanitized.
 		sanitize_html( $message )


afragen:
So why does it say `Validated as gitupdater.com`?  I do have that domain registered and it redirects to git-updater.com but in this usage where does it get the value being validated?

toderash:
@afragen these may be mockups rather than screenshots (not sure).

@rmccue these make sense to me, but the missing state could cause confusion - it looks invalid. (The image loaded once for me and not gives a 404, so going from memory.)

I understand the 3 states to be:

- domain.com is a valid alias for the package DID
- domain.com was supplied as an alias for the package DID but we could not validate it
- no domain alias was provided for the package DID

We just need to say it in fewer words that convey those meanings. Does this capture the _intended meanings_ as you see it? If so, we can work at making it concise without changing the meaning.

rmccue:
To be clear, the status here is mocked to demonstrate the design.

> We just need to say it in fewer words that convey those meanings. Does this capture the _intended meanings_ as you see it? If so, we can work at making it concise without changing the meaning.

Yep, basically.

The current output here specifically is:
* valid: "**Validated** as {domain}" (with a link to said domain)
* invalid: "**Error:** Failed domain alias validation, this package may be unsafe: {validation error}"
* missing: "Not validated: No domain alias is set"

The error state also adds a big ol' red error message to make errors even clearer.

We _could_ show nothing at all if there's no domain alias set - I considered this, but figured I'd add an explicit message at least for now to discuss.

toderash:
How about something like these?
mockup-did-domain-validation
Not totally committed to them, but
- missing alias is reported in neutral language
- redirection note added
On the latter point, I'm thinking it would be common to have the domain but redirect it to Github (for example) so if the user clicks on the link, they'll land somewhere unexpected based on the validation method and link text.

Where validation fails, we could potentially show a limited set of reasons, like "DNS does not resolve" or something - or just add a verbal shrug for "We don't know, maybe check again later."

Not committed to the preposition here either - "at" "by" "to" "as" are all true, with slightly different nuances. _at_ the domain (DID lives here), _by_ the domain (not a third-party validation), _to_ the domain (where the DID points), _as_ (because AKA). Could also be "Resolves to" or "as".

Some combination there should work since it leave the interpretation to the user as to the message's implications.

inc/packages/namespace.php
```diff
@@ -724,4 +725,106 @@ function maybe_add_accept_header( $args, $url ) : array {
...
+ * Uses `fair://` aliases from the DID document to determine the alias for the
+ * package. Performs bidirectional validation using DNS to ensure the DID is
+ * valid for the given alias.
+ *
+ * Uses cached result for one hour.
+ *
+ * @return string|WP_Error|null Alias domain if successfully validated, null if no valid alias is set, or error otherwise.
+ */
+function validate_package_alias( DIDDocument $did ) {
+	$cache_key = sprintf( 'fair_did_alias_' . $did->id );
```
afragen: Why use `sprintf`?

rmccue: whoops, should have been `sprintf( 'fair_did_alias_%s', $did->id )`

inc/packages/admin/info.php
```diff
@@ -430,6 +435,70 @@ function add_requirement_notices( ReleaseDocument $release ) : void {
...
+
+		case 'NULL':
+			$message = __( 'Not validated: No domain alias is set', 'fair' );
+			$result = true;
+			break;
+
+		default:
+			if ( ! is_wp_error( $validation ) ) {
+				// Invalid type, assume failure.
+				$validation = new WP_Error( 'fair.packages.admin.info.validation_notice.invalid_result', 'An unknown error occurred' );
```
costdev: `WP_Error` messages elsewhere in the plugin are translated, but I don't think any of those in this PR are translated.

rmccue:
Per Slack discussion, we tweaked slightly to add "Domain Alias" as a title, but otherwise are sticking with the microcopy as noted previously.

Ready for final review.

inc/packages/admin/info.php
```diff
@@ -430,6 +435,76 @@ function add_requirement_notices( ReleaseDocument $release ) : void {
...
+						'paragraph_wrap'     => false,
+					]
+				);
+			} );
+			$result = false;
+			break;
+	}
+
+	printf(
+		'<strong>%s</strong>: %s',
```
costdev: Technically, languages like French could have something like `<strong>%s</strong> : %s` - Note the space after the colon. That's a simple example, but shows we might want this string to be translatable too in case others use different formats.

rmccue: Oh, it's actually just straight wrong anyway, the colon's meant to be in the title.

inc/packages/namespace.php
```diff
@@ -724,4 +725,116 @@ function maybe_add_accept_header( $args, $url ) : array {
...
+ * valid for the given alias.
+ *
+ * Uses cached result for one hour.
+ *
+ * @param DIDDocument $did DID to validate.
+ * @return string|WP_Error|null Alias domain if successfully validated, null if no valid alias is set, or error otherwise.
+ */
+function validate_package_alias( DIDDocument $did ) {
+	$cache_key = sprintf( 'fair_did_alias_%s', $did->id );
+	$cached = get_site_transient( $cache_key );
```
afragen, line 742: I believe this is the only place we use `set_site_transient()` and `get_site_transient()`. All other instances are `get_transient()` and `set_transient()`.

afragen (approved):
My only question is re: site transients.

inc/packages/namespace.php
rmccue: Might be worth us considering using site transients for some of those in a follow-up? I used the site transients here so that multiple sites can benefit from the caching, whereas things like hitting the aggregator could be altered on a per-site basis (eg filtering to add params, etc).

rmccue: https://github.com/fairpm/fair-plugin/issues/257

PR #244, opened 2025-09-02
Icons: Guard against a non-array updates response

afragen:
For some reason I saw a number of the following errors. It seems to happen after the cache is cleared.

`Warning: foreach() argument must be of type array|object, null given in /sites/thefragens.com/files/wp-content/plugins/fair-plugin/inc/icons/namespace.php on line 38`

It seems that sometimes `$transient->response` can be null. I have no idea why but this should fix it.

### DIFF
--- a/inc/icons/namespace.php
+++ b/inc/icons/namespace.php
@@ -35,7 +35,7 @@ function set_default_icon( $transient ) {
 		return $transient;
 	}
 
-	foreach ( $transient->response as $updates ) {
+	foreach ( (array) $transient->response as $updates ) {
 		$url = plugin_dir_url( PLUGIN_FILE ) . 'inc/icons/svg.php';
 		$url = add_query_arg( 'color', set_random_color(), $url );
 		$updates->icons['default'] = $url;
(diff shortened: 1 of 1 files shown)


inc/icons/namespace.php
```diff
@@ -35,7 +35,7 @@ function set_default_icon( $transient ) {
 		return $transient;
 	}
 
-	foreach ( $transient->response as $updates ) {
+	foreach ( (array) $transient->response as $updates ) {
```
costdev: It's more robust to add this to the guard above to avoid gotchas in casting:

```php
|| ! is_array( $transient->response )
```

e.g. Say for some reason `$transient->response` is some other unexpected value like `false`. Being a transient means that's quite easily possible. With the guard, the function will bail - safe and sound. With casting, the function will do `(array) false` - which would become `[ 0 => false ]` and subsequently error on `$updates->icons['default']`. Bailing early on a non-array reduces the probability of things like that.

costdev (approved):

PR #246, opened 2025-09-08
Prevent pings of unpublished URLs.

peterwilsoncc:
Follow up to #160.

~In my earlier pull request, I _broke compatibility for WordPress versions prior to 5.7.0_ as `is_post_status_viewable()` was unavailable in those versions. I've introduced a polyfill to use in it's place.~

The previous pull request was sending the permalink if the new status was unviewable and the old status was. This was another mistake by yours truely. In this case WP would submit the plain permalink used for previewing links, eg `example.com?p=1`, rather than the original link sent to IndexNow, eg `example.com/hello-world`

~ping @rmccue @afragen: this is probably important given the fatal error I introduced in old WP versions.~

### DIFF
--- a/inc/pings/namespace.php
+++ b/inc/pings/namespace.php
@@ -102,42 +102,6 @@ function handle_key_file_request() {
 }
 
 /**
- * Determines whether a post status is considered "viewable".
- *
- * Polyfill of `is_post_status_viewable()` for WordPress versions prior to 5.7.0.
- *
- * @param string|stdClass $post_status Post status name or object.
- * @return bool Whether the post status should be considered viewable.
- */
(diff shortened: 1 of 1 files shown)


afragen:
Might it not be better to put the polyfill in `inc/compatibility/wp-polyfill.php`?

peterwilsoncc:
> Might it not be better to put the polyfill in `inc/compatibility/wp-polyfill.php`?

Fixed in c2de3c689322990574c341b62c1624e94fa3797a -- it was already polyfilled

afragen (approved):

afragen:
I'm pretty sure you need to sign on if the commits. Signing and a force push works.

Issue #231, opened 2025-08-23
Missing signature verification for did:plc:... packages


kasparsd:
It appears that the plugin currently doesn't implement any kind of signature verification for packages installed via `FairPackageManagementRepo` (or other repos). I couldn't find this described in the README or any open issues/pull-requests that would implement it.

Even the FAIR plugin itself doesn't use a did:plc:... identifier or any other mechanism for validating the install ZIPs. 

It attempts to check that the signing keys are defined but doesn't actually use them anywhere:

https://github.com/fairpm/fair-plugin/blob/3bef1caf08375b9061dcb085da10d2aa071df934/inc/packages/namespace.php#L236-L239

Is this part of the current roadmap?

kasparsd:
Here is some prior work that was excluded from previous releases because of other priorities:

https://github.com/fairpm/fair-plugin/blob/7954c09458d5bdba75351a97d7e4496337481665/inc/packages/class-upgrader.php#L197-L211

afragen:
I think we can close this issue. Please re-open if I am wrong.

PR #247, opened 2025-09-09
Verify signatures when downloading packages

costdev:
See #231 

This PR introduces signature verification for packages. The PR still needs initial testing.

### Approach

The `'upgrader_pre_download'` hook is used to:
- Download the package
- Set the trusted keys
- Perform signature verification

The callback includes guards and clean up surrounding the steps above.

Using this particular hook means minimal duplication of Core's code while returning back into Core's process as soon as possible.

### New dependency
This PR introduces [yocLib - Multibase](https://github.com/yocto/yoclib-multibase-php) as a dependency. This is used to convert signing keys from Base58BTC to Base64. While we could use another implementation, this is used elsewhere in the project and I erred on the side of getting something testable out quicker.

### Pre-merge note
If we continue with this dependency, or find another external dependency, the bundler script will need to be checked to ensure it includes/excludes files as its supposed to. This should be done before this PR is merged.


composer.json
```diff
@@ -9,7 +9,8 @@
         }
     ],
     "require": {
-        "php": ">=7.4"
+        "php": ">=7.4",
+        "yocto/yoclib-multibase": "^1.2"
```
kasparsd: This appears to be the first external dependency for this plugin. Should we consider isolating the namespace so that it avoids conflicts and class redefine errors if any other package is depending on the same dependency?

This is a client-facing plugin that will be running on every possible setup of WP plugins and dependencies. Maybe consider the approach used [by Sitekit](https://github.com/google/site-kit-wp/blob/92d8456beba77358c41b3d60afd298be4d26158f/scoper.inc.php) and [Yoast](https://github.com/Yoast/wordpress-seo/blob/85a0f13e0a5a5fc94da32e3a61157e485f895c4e/composer.json#L151-L163)?

Alternatively, could we fork the relevant parts if that's only part of the library that we're using?

inc/updater/class-updater.php
```diff
@@ -129,9 +130,86 @@ public function load_hooks() {
...
+		if ( ! preg_match( '!^(http|https|ftp)://!i', $package ) && file_exists( $package ) ) {
+			return $package;
+		}
+
+		$artifact = Packages\pick_artifact_by_lang( $this->release->artifacts->package );
+		if ( ! $artifact || $package !== $artifact->url ) {
+			return $reply;
+		}
+
+		$path = $upgrader->download_package( $package, false, $hook_extra );
```
kasparsd, line 184: Just to confirm -- this method [supports signature verification](https://github.com/WordPress/wordpress-develop/blob/c726220a21d13fdb5409372b652c9460c59ce1db/src/wp-admin/includes/class-wp-upgrader.php#L335) but we're not using that because it would:

1. require filtering `wp_signature_hosts`
2. and relying on the `X-Content-Signature` response header of the package ZIP for the signature.

However, we're fetching the signature from the package manifest JSON so we must use our own verification flow, right?

inc/updater/class-updater.php
```diff
@@ -129,9 +130,86 @@ public function load_hooks() {
...
+		if ( is_wp_error( $path ) ) {
+			return $path;
+		}
+
+		add_filter( 'wp_trusted_keys', [ $this, 'get_trusted_keys' ], 100 );
+		$result = verify_file_signature( $path, $artifact->signature );
+		remove_filter( 'wp_trusted_keys', [ $this, 'get_trusted_keys' ], 100 );
+
+		return $result;
```
kasparsd: Should we return the `$path` here if the verification passes? We're currently returning boolean `false` if the verification fails [which would make the WP core download it anyway](https://github.com/WordPress/wordpress-develop/blob/c726220a21d13fdb5409372b652c9460c59ce1db/src/wp-admin/includes/class-wp-upgrader.php#L321-L323).

I believe this filter should either return the downloaded file path on signature verification success or an `WP_Error` if it fails.

costdev: 1. Should we return the `$path` here if verification passes?
    - Yes, yes we should. Woops!
    - I'll update that return statement to `return is_wp_error( $result ) ? $result : $path;`
2. We're currently returning boolean false if the verification fails
    - While `verify_file_signature()` is documented to return "True on success, false if verification not attempted, or WP_Error describing an error condition.", it actually only ever returns `true|WP_Error`. [Take a look](https://github.com/WordPress/wordpress-develop/blob/6.8.2/src/wp-admin/includes/file.php#L1405-L1531).

inc/updater/class-updater.php
costdev: Exactly. 👍

composer.json
costdev: I'm totally happy for us to handle the conversion of Base58BTC -> Base64 any way we want. The less we have to maintain (whether internal or external), the better.

kasparsd:
I'm thinking about the alternative workflows where this could apply. Will this apply to the `did:` installer too? Or if a user goes through the manual install workflow to upload _an update_ to an existing package?

inc/updater/class-updater.php
costdev: > I'll update that return statement to return `is_wp_error( $result ) ? $result : $path;`

Done in: [d5be080](https://github.com/fairpm/fair-plugin/pull/247/commits/d5be080d4bddee027b7195cd0d874ec69cfe802f).

composer.json
kasparsd: Indeed! I believe the same thing needs to be solved for the mini fair so having a consistent solution would help too.

The implementation [appears to be ~40 lines](https://github.com/yocto/yoclib-multibase-php/blob/c7171897bf61dbc4a4cc6bb3f2fd5c3e62298e13/src/Base58.php#L12-L60) so could we copy-paste it here with the appropriate attribution?

costdev:
> Will this apply to the `did:` installer too?

Yeah this should work with the `did:` direct installer right away.

> Or if a user goes through the manual install workflow to upload _an update_ to an existing package?

Yes I believe so. The FAIR plugin [sets up the Updater](https://github.com/fairpm/fair-plugin/blob/689c728237049c145bce620a164e4d5bd459f883/inc/updater/namespace.php#L62) for each installed package on [init](https://github.com/fairpm/fair-plugin/blob/main/inc/updater/namespace.php#L14). The `Updater::run()` method calls `Updater::load_hooks()`, which is where this PR hooks to `upgrader_pre_download` for that package.

I'm not sure whether this is the case when uploading _a new_ package though. I expect that is different.

composer.json
costdev: Yep totally makes sense. In case there are differing opinions on that, I'll hold off for a bit on making that change to the PR until others have had a chance to weigh in.

inc/updater/class-updater.php
```diff
@@ -129,9 +130,86 @@ public function load_hooks() {
...
+		$path = $upgrader->download_package( $package, false, $hook_extra );
+		if ( is_wp_error( $path ) ) {
+			return $path;
+		}
+
+		add_filter( 'wp_trusted_keys', [ $this, 'get_trusted_keys' ], 100 );
+		$result = verify_file_signature( $path, $artifact->signature );
+		remove_filter( 'wp_trusted_keys', [ $this, 'get_trusted_keys' ], 100 );
+
+		return is_wp_error( $result ) ? $result : $path;
```
kasparsd: You correctly pointed out that `verify_file_signature()` currently only returns `true` or `WP_Error` but that could change over time.

Could we make this very explicit:


```php
if ( true === $result ) { // This is the only condition we accept as passing.
	return $path;
} elseif ( is_wp_error( $result ) ) {
	return $result;
}

return new WP_Error( '...' ); // Our own error of signature verification failure.
```

kasparsd:
> I'm not sure whether this is the case when uploading a new package though. I expect that is different.

Yeah, there is no reliable way to determine the package vendor if we only have the ZIP. For a secure installation users should use the `did:` installer.

We could attempt to extract the vendor `did` from the uploaded package meta header and then compare the downloaded file hash/checksum + signature with any of the available releases. But that doesn't really help because a malicious actor could obviously remove the `did` info from the meta.

I love the did installer! Such a simple UI for a super-secure installer.

composer.json
rmccue: imo, we should import the implementation here; the upstream dep is a generic implementation of multibase, but all our flows only ever use base58-btc.

Note that the linked implementation handles the base conversion, but multibase also includes the prefix ([`z` for base58-btc](https://github.com/yocto/yoclib-multibase-php/blob/c7171897bf61dbc4a4cc6bb3f2fd5c3e62298e13/src/Multibase.php#L26)), so if you port it over make sure you're checking and stripping that too.

costdev:
FYI: After chatting with @rmccue, I'm going to rework some of this to use a subclass of the `WP_Upgrader` class so we have some more control and it'll read better.

inc/updater/class-updater.php
costdev: Although I'll be changing the implementation, I've gone ahead and done this in [7da9b35](https://github.com/fairpm/fair-plugin/pull/247/commits/7da9b3591fd34695ce6421b7533959d9ee90c15e) so I don't forget to include it in the alternative implementation.

composer.json
costdev: Done in [319d7a0](https://github.com/fairpm/fair-plugin/pull/247/commits/319d7a0a8eb2430868da3ff9d7dda0958dddab0c), including the `'z'` prefix being stripped in `Base58BTC::decode()`.

inc/updater/class-base58btc.php
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+namespace FAIR\Updater;
+
+/**
+ * Base58BTC encoding and decoding.
+ */
+class Base58BTC {
+	const PREFIX = 'z';
+	const ALPHABET = '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz';
```
afragen, line 16: No zero?

costdev: Nope:
- [yocLib - Multibase](https://github.com/yocto/yoclib-multibase-php/blob/c7171897bf61dbc4a4cc6bb3f2fd5c3e62298e13/src/Multibase.php#L41)
- [LearnMeABitcoin](https://learnmeabitcoin.com/technical/keys/base58/)
- [Bitcoin 0.1](https://github.com/0xMagnuz/Bitcoin-v0.1/blob/2c7a4c9bbb793c5029fc9363d504944b62f1ec4d/bitcoin0.1/src/base58.h#L8) (and the alphabet at ~Line 16)

Base58 excludes these:
1.  `0` (zero)
2. `O` (uppercase o)
3. `l` (lowercase L)
4. `I` (uppercase i)
5. `+` (plus)
6. `/` (forward slash)

costdev:
The force-push was a rebase on `main` due to a merge conflict.

rmccue:
In https://github.com/fairpm/mini-fair-repo/pull/52 we switched the signature algorithm to sha384, which I think was the last blocker here. @costdev is this now ready for review/merge?

costdev:
@rmccue Yes, ready for review, testing and merge. At the moment, you can test by enabling signature verification and installing Git Updater via Direct Install - `did:plc:afjf7gsjzsqmgc7dlhb553mv`.

MU Plugin to enable signature verification:
```php
<?php
/**
 * Plugin Name: FAIR: Verify Package Signatures
 * Description: Enables FAIR package signature verification
 * Author:      FAIR Contributors
 * License:     GPLv2 or later
 * Version:     1.0.0
 */
add_filter( 'fair.packages.updater.verify_signatures', '__return_true' );
```

costdev:
I've added an `error_log()` so testers can see when it's successfully verified the package. Just ensure `WP_DEBUG` and `WP_DEBUG_LOG` are set to `true`.

**DO NOT MERGE WITH THIS `error_log()` in the PR - I'll remove it once the PR has 1-2 approvals **

afragen:
I now see this issue

screenshot_717

afragen:
If the package is not yet installed then `Updater\get_packages()` returns `null`

costdev:
That's related to another PR. I'll get that fixed shortly.

afragen:
screenshot_718

afragen:
Cache issue. All working/verifying.

inc/packages/admin/namespace.php
```diff
@@ -90,6 +90,8 @@ function handle_did_during_ajax( $result, $action, $args ) {
 		return $result;
 	}
 
+	( new Updater\Updater( $did ) )->run();
```
rmccue, line 93: For the future: we should rename this method to be clearer, and move some of the global state checks out of it.

inc/updater/class-updater.php
```diff
@@ -124,14 +125,128 @@ public function run() {
...
+	public function verify_signature_on_download( $reply, string $package, WP_Upgrader $upgrader, $hook_extra ) {
+		static $has_run = [];
+
+		if ( false !== $reply || ( ! $upgrader instanceof Plugin_Upgrader && ! $upgrader instanceof Theme_Upgrader ) ) {
+			return $reply;
+		}
+
+		// This method is hooked to 'upgrader_pre_download', which is used in WP_Upgrader::download_package().
+		// Bailing on subsequent runs for the same package URI prevents an infinite loop.
+		$key = md5( $this->did . '_' . $package );
```
rmccue: Do we need to use md5 here? If it's just for a general hash, we should use `sha1()` - `md5()` gets flagged by loads of automated scanners.

costdev: Done in [8b849db](https://github.com/fairpm/fair-plugin/pull/247/commits/8b849dbdf1950287b6de6d3bdf6bd6ee9107e4b6).

rmccue (approved):

PR #248, opened 2025-09-09
Change from wp_cache_* to *_transient for fair-plugin

afragen:
Use transients not wp_cache_* for fair-plugin

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -7,17 +7,19 @@
 
 namespace FAIR\Packages;
 
+use const FAIR\CACHE_BASE;
+use const FAIR\CACHE_LIFETIME;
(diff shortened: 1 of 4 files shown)


inc/packages/namespace.php
```diff
@@ -99,7 +99,7 @@ function get_did_document( string $id ) {
 	if ( is_wp_error( $document ) ) {
 		return $document;
 	}
-	wp_cache_set( $id, $document, 'did-docs', CACHE_LIFETIME );
+	set_transient( $id, $document, CACHE_LIFETIME );
```
costdev: Should the transient names be prefixed since we don't have the cache group anymore?

afragen:
Seems like something for another PR unless you want me to update here?

costdev (approved):

PR #250, opened 2025-09-20
Use bridged data from legacy endpoints

rmccue:
* Adds _fair parameter to requests to the repository (except .org if that's being used)
* Uses the FAIR info page instead of default, if there's valid bridged data
* Displays the repository name on the info page and card for bridged packages

Screenshot 2025-09-20 at 21 44 02@2x

Screenshot 2025-09-20 at 21 44 27@2x

### DIFF
--- a/inc/packages/admin/info.php
+++ b/inc/packages/admin/info.php
@@ -280,6 +280,9 @@ function render_fyi( MetadataDocument $doc, ReleaseDocument $release ) : void {
 			<?php if ( ! empty( $doc->slug ) ) : ?>
 				<li><strong><?= __( 'Slug:', 'fair' ); ?></strong> <?= esc_attr( $doc->slug ); ?></li>
 			<?php endif; ?>
+			<?php if ( ! empty( $doc->id ) ) : ?>
+				<li><strong><?= __( 'ID:', 'fair' ); ?></strong> <code><?= esc_attr( $doc->id ); ?></code></li>
+			<?php endif; ?>
 			<?php if ( ! empty( $release->requires ) ) : ?>
 				<li>
 					<strong><?= __( 'Requires:', 'fair' ); ?></strong>
@@ -322,8 +325,36 @@ function render_fyi( MetadataDocument $doc, ReleaseDocument $release ) : void {
 				?>
 			</ul>
 		<?php endif ?>
+		<?php
+		$repo_host = get_repository_hostname( $doc->id );
+		if ( $repo_host ) :
+			?>
+			<p><small>Plugin available via FAIR repository hosted at <?php echo esc_html( $repo_host ); ?></small></p>
+		<?php else: ?>
+			<p><small>Plugin available via FAIR repository</small></p>
+		<?php endif; ?>
 	</div>
 	<?php
+}
+
+/**
+ * Get the hostname for the repository hosting a package.
+ *
+ * @param string $did DID to check.
+ * @return string|null Hostname if available, null if there's an error.
+ */
+function get_repository_hostname( string $did ) : ?string {
+	$did_doc = Packages\get_did_document( $did );
+	if ( is_wp_error( $did_doc ) ) {
+		return null;
+	}
+
+	$repo = $did_doc->get_service( Packages\SERVICE_ID );
+	if ( empty( $repo ) ) {
+		return null;
+	}
+
+	return parse_url( $repo->serviceEndpoint, PHP_URL_HOST );
 }
 
 /**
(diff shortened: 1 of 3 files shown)


inc/packages/admin/info.php
```diff
@@ -322,10 +325,38 @@ function render_fyi( MetadataDocument $doc, ReleaseDocument $release ) : void {
...
+	if ( is_wp_error( $did_doc ) ) {
+		return null;
+	}
+
+	$repo = $did_doc->get_service( Packages\SERVICE_ID );
+	if ( empty( $repo ) ) {
+		return null;
+	}
+
+	return parse_url( $repo->serviceEndpoint, PHP_URL_HOST );
```
costdev: In the event of a malformed URL, `parse_url()` will return `false`. Should we handle this in case the DID document (and serviceEndpoint) is created by something other than Mini FAIR repo?

inc/packages/admin/namespace.php
```diff
@@ -271,3 +298,59 @@ function maybe_hijack_plugin_info() {
...
+		return $description;
+	}
+
+	$did = $plugin['_fair']['id'];
+	$repo_host = Info\get_repository_hostname( $did );
+	if ( empty( $repo_host ) ) {
+		return $description;
+	}
+
+	$description .= '</p><p class="authors"><em>' . sprintf( __( 'Hosted on %s', 'fair' ), esc_html( $repo_host ) ) . '</em>';
```
costdev: Can we drop a translators comment here so it's clear `%s` isn't, for example, the Repository Host's name?

Just thinking about `Hosted on fair.git-updater.com` vs `Hosted on Git Updater`, and how that could affect a translation, such as the appropriate choice between on/at/from/by/with/etc. for a given locale.

rmccue:
AC has been updated to return the right ID, so this is now functional; using AC staging with the constant above, you should be able to search for and find Git Updater to trigger this.

inc/packages/admin/info.php
rmccue: I don't think it'd ever end up here (since you wouldn't be able to actually load the metadata if the DID was malformed), but can do.

rmccue:
Installation via the iframe is working, but via the button in the card isn't, as we need to switch the slug to our temporary fake one (`git-updater-did--plc--ajf...`). When we do this, the button does not update its status - the JS code in core is looking for `plugin-card-{slug}`, but because we're altering the slug, that's breaking.

Filtering `plugins_api_result` to change the slugs solves this, but when it calls `check_plugin_dependencies`, _that_ returns the permanent slug (`git-updater-f27e06`) - which then doesn't match the class name 😅

rmccue:
Worked around via a subclass with output buffering.

Ideally, we would override individual cards - sadly, the core list table has the loop iterations _inline_ rather than in a separate method, meaning we can't override them individually.

inc/packages/admin/namespace.php
```diff
@@ -36,6 +36,24 @@ function bootstrap() {
...
+ *
+ * @param string $class_name List table class name to use.
+ * @return string Overridden class name.
+ */
+function maybe_override_list_table( $class_name ) {
+	if ( $class_name !== 'WP_Plugin_Install_List_Table' ) {
+		return $class_name;
+	}
+
+	// Load list table class. (We must do this here, as WP_List_Table isn't loaded by default.)
```
costdev: Is this confirmed?

Looking at [these lines](https://github.com/WordPress/wordpress-develop/blob/6.8.2/src/wp-admin/includes/list-table.php#L47-L50), it looks like the parent class should be loaded before the `wp_list_table_class_name` filters are applied.

It does no harm to have the `require_once` anyway. I was just curious if you got errors before doing so.

rmccue: I'd copied it across from when I was loading it separately, but I figured no harm no foul.

costdev (approved):
Looks good to me.

We have two outstanding paths:
1. Install from card -> open "More Details" -> Activate button doesn't show.
    - This is due to us not having implemented the "Activate" button yet in `get_action_button()`. I'll open a PR soon after we merge this one.
    - IMO, this isn't a blocker for 1.0.0.
    - ✅ Update - This path is resolved:
        - #251
        - #253
3. Install from "More Details" -> Activate button doesn't show in the card behind.
    - This is due to a mismatch between the `response.slug` from the plugin dependencies check, and the `data-slug` attribute on the link. The mismatch is intentional, and will require a change of the overall approach.
    - IMO, this isn't a blocker for 1.0.0.

afragen:
Seems to work great!

PR #258, opened 2025-09-22
Allow searching by DID

costdev:
This PR enables searching by DID on the `Plugins > Add New Plugin` screen.

### DIFF
--- a/inc/packages/admin/namespace.php
+++ b/inc/packages/admin/namespace.php
@@ -28,6 +28,7 @@ function bootstrap() {
 
 	add_filter( 'install_plugins_tabs', __NAMESPACE__ . '\\add_direct_tab' );
 	add_filter( 'plugins_api', __NAMESPACE__ . '\\handle_did_during_ajax', 10, 3 );
+	add_filter( 'plugins_api', __NAMESPACE__ . '\\search_by_did', 10, 3 );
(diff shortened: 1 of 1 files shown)


costdev:
Note: The `last_updated` needs data from the metadata endpoint. Can we add this to package artifacts in ReleaseDocuments, or something else?

afragen:
Agreed, last_updated would be nice, but wouldn't that involve a spec change?

afragen:
Other question. Can't the DID be pulled into the metadata on AC import and then use that as a search parameter? Bonus is we get all the other data elements as well. 

cc: @chuckadams

costdev:
> Agreed, last_updated would be nice, but wouldn't that involve a spec change?

It would, but this feature sort of reveals it as a missing element. The number of installs and ratings fall under stats, which are still in our roadmap, but the `last_updated` seems like it should absolutely be part of an artifact's data. That means the `last_updated` becomes independent of a repository, which is important if a plugin/theme author decides to switch repository at some point.

costdev:
> Other question. Can't the DID be pulled into the metadata on AC import and then use that as a search parameter? Bonus is we get all the other data elements as well.

It can but IMO if a user is entering a DID, it makes sense to communicate with the DID's registered metadata source.

costdev:
A `last_updated` value from the metadata source depends on the outcome of this issue: https://github.com/fairpm/fair-protocol/issues/51

PR #277, opened 2025-09-29
Add WP-CLI support

costdev:
This PR adds support for the following `wp plugin <subcommand> <did>` commands.

The following commands are supported:
- `activate`
- `deactivate`
- `delete`
- `get`
- `install`
- `is-active`
- `path`
- `search`
- `status`
- `toggle`
- `uninstall`
- `update`

The following commands are not currently supported:
- `verify-checksums` - This relies on the wordpress.org API, and each file in a package has its own checksum which isn't part of the FAIR protocol.

The PR also includes refactoring some of the plugin to facilitate WP-CLI support:
- How a DID for install is cached: Previously, the DID was cached in the AJAX installation handler. This meant that only one DID could be cached, even during a bulk install (such as via WP-CLI). Now, the DID is cached when its artifact's URL matches the package currently being passed for installation. Additionally, the cached DID is cleared after the package has been installed to prevent contamination.
- Signature verification functions: Previously, signature verification was only applied in the Updater class, whose setup is somewhat tightly coupled to installing in an admin context. Now, signature verification has been moved to general functions so that it can be used in various contexts.
- Search by DID: This has been moved from the `Packages\Admin` namespace to the `Packages` namespace, as this is now used in more than just the admin context.

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -19,6 +19,7 @@ use WP_Upgrader;
 const CACHE_KEY = CACHE_BASE . 'packages-';
 const CACHE_METADATA_DOCUMENTS = CACHE_BASE . 'metadata-documents-';
 const CACHE_RELEASE_PACKAGES = CACHE_BASE . 'release-packages';
+const CACHE_DID_FOR_INSTALL = 'fair-install-did';
 const CONTENT_TYPE = 'application/json+fair';
 const SERVICE_ID = 'FairPackageManagementRepo';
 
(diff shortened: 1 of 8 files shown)


costdev:
Some DIDs to test with:
- Git Updater: `did:plc:afjf7gsjzsqmgc7dlhb553mv`
- Handbook Callout Blocks: `did:plc:deoui6ztyx6paqajconl67rz`
- Pods: `did:plc:e3rm6t7cspgpzaf47kn3nnsl`

afragen (approved):

inc/packages/namespace.php
```diff
@@ -837,4 +874,96 @@ function fetch_and_validate_package_alias( DIDDocument $did ) {
...
+	$api_data['rating'] ??= 0;
+	$api_data['active_installs'] ??= 0;
+
+	// Avoid a double-hashed slug.
+	$hash_suffix = '-' . get_did_hash( $did );
+	if ( str_ends_with( $api_data['slug'], $hash_suffix ) ) {
+		$api_data['slug'] = str_replace( $hash_suffix, '', $api_data['slug'] );
+	}
+
+	return $api_data;
```
afragen, line 940: Should we rename get_update_data to something like get_package_data() and just add 

- description
- short_description
- last_updated
- num_ratings
- rating
- active_installs

We could then simplify this function.

costdev: Yeah agreed - though if we moved those to `get_update_data()`, the only real difference is that the didhash gets stripped from the slug here. The only reason that's done is because somewhere along the line, the didhash is added to the slug again, so it was becoming `slug-didhash-didhash`. IIRC it was happening in the plugin cards (?) but I couldn't find what was adding it at the time.

If we can locate where that gets done, we can make sure it only adds the didhash if it's not already there, then we can actually remove `get_api_data()` altogether

While not a blocker for this PR, I'll see if I can track that down.

afragen: Or we could add a `slug_didhash` to the data returned from the new `get_package_data()` then just use that as needed.

costdev: Yeah we could add the slug_didhash too so it's available without needing to call get_did_hash() in each location, though I'd also be happy enough if we just had slug and hashed it whenever needed. No strong feelings one way or the other.

costdev: Let's follow this up after this PR is merged since we'll need to look into the areas where we do and don't need the hashed slug.

swissspidy:
@costdev If there's anything we could improve on the WP-CLI side to make this more seamless, please let me know :-)

PR #286, opened 2025-10-04
Add more data to get_package_data(), renamed from get_update_data()

afragen:
Rename `get_update_data()` to `get_package_data()`.

Add data elements needed for `plugins_api` or `themes_api` filter.

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -579,7 +579,7 @@ function get_hashed_filename( $metadata ) : string {
  * @param string $did DID.
  * @return array|WP_Error
  */
-function get_update_data( $did ) {
+function get_package_data( $did ) {
 	$metadata = fetch_package_metadata( $did );
 	if ( is_wp_error( $metadata ) ) {
 		return $metadata;
@@ -593,16 +593,20 @@ function get_update_data( $did ) {
 	$required_versions = version_requirements( $release );
 	$filename = get_hashed_filename( $metadata );
 	$type = str_replace( 'wp-', '', $metadata->type );
+	$sections = (array) $metadata->sections;
 
 	$response = [
 		'name'             => $metadata->name,
 		'author'           => $metadata->authors[0]->name,
 		'author_uri'       => $metadata->authors[0]->url,
-		'slug'             => $metadata->slug . '-' . get_did_hash( $did ),
+		'slug'             => $metadata->slug,
+		'slug_didhash'     => $metadata->slug . '-' . get_did_hash( $did ),
 		$type              => $filename,
 		'file'             => $filename,
 		'url'              => $metadata->url ?? $metadata->slug,
-		'sections'         => (array) $metadata->sections,
+		'sections'         => $sections,
+		'description' => $sections['description'] ?? '',
+		'short_description' => substr( strip_tags( trim( $sections['description'] ) ), 0, 147 ) . '...',
 		'icons'            => isset( $release->artifacts->icon ) ? get_icons( $release->artifacts->icon ) : [],
 		'banners'          => isset( $release->artifacts->banner ) ? get_banners( $release->artifacts->banner ) : [],
 		'update-supported' => true,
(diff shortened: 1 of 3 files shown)


costdev:
Looks like there's a breakage in the PR:

1. Go to Plugins > Add New.
2. Install a plugin by DID, for example, `did:plc:e3rm6t7cspgpzaf47kn3nnsl`.
3. Unbump its version.
4. Go back to Plugins > Add New and search for the DID again.
5. 🐛 The button says "Install Now" rather than "Update Now".

inc/packages/namespace.php
```diff
@@ -593,29 +593,37 @@ function get_update_data( $did ) {
...
+		'author'            => $metadata->authors[0]->name,
+		'author_uri'        => $metadata->authors[0]->url,
+		'slug'              => $metadata->slug,
+		'slug_didhash'      => $metadata->slug . '-' . get_did_hash( $did ),
+		$type               => $filename,
+		'file'              => $filename,
+		'url'               => $metadata->url ?? $metadata->slug,
+		'sections'          => $sections,
+		'description'       => $sections['description'] ?? '',
+		'short_description' => substr( strip_tags( trim( $sections['description'] ) ), 0, 147 ) . '...',
```
costdev: I'm just noticing a couple of things:

1. We have a null check on `$sections['description']` above but not when setting `short_description`.
2.  If `$sections['description']` is `''` (whether before or after trimming), the `short_description` will be `'...'`.

If we set a local `$description` variable above the `$response` definition:
```php
$description = trim( $sections['description'] ?? '' );
```

Then we can just set `description` and `short_description` like so:
```php
'description' => $description,
'short_description' => $description ? substr( strip_tags( $description ), 0, 147 ) . '...' : '';
```

That will fix both issues above.

We can sort that in a follow-up PR though since this PR didn't introduce them and it's not particularly scope for this PR.

afragen:
Can confirm but this is only in the plugin card. Need to look into plugin card and button creation.

inc/updater/class-updater.php
```diff
@@ -237,7 +237,7 @@ public function update_site_transient( $transient ) {
 
 		$rel_path = plugin_basename( $this->filepath );
 		$rel_path = 'theme' === $this->type ? dirname( $rel_path ) : $rel_path;
-		$response = Packages\get_update_data( $this->did );
+		$response = Packages\get_package_data( $this->did );
 		if ( is_wp_error( $response ) ) {
 			return $transient;
 		}
```
costdev, line 243: Looks like the issue is here. Need to swap the `slug` for `slug_didhash`.

```php
$response = Packages\get_package_data( $this->did );
if ( is_wp_error( $response ) ) {
	return $transient;
}
$response['slug'] = $response['slug_didhash'];
```

inc/packages/namespace.php
afragen: Why do we need to set a new variable, `$description`?

inc/updater/class-updater.php
afragen: I think I found the issue in `alter_slugs()`. Need to exit if plugin already installed.

inc/packages/namespace.php
costdev: Just so that the `trim()` and `?? ''` only need to be done once

afragen:
@costdev your 🐛 seems to be somewhere in `alter_slugs` and `maybe_hijack_install_button`

costdev:
It's a bit weird, because when I did [this](https://github.com/fairpm/fair-plugin/pull/286#discussion_r2404211843), the "Update now" button appeared as expected. Not saying that's the _right_ approach, but that data passes on to the determination of what the status is, or which button to show.

inc/packages/namespace.php
```diff
@@ -593,29 +593,38 @@ function get_update_data( $did ) {
 	$required_versions = version_requirements( $release );
 	$filename = get_hashed_filename( $metadata );
 	$type = str_replace( 'wp-', '', $metadata->type );
+	$sections = (array) $metadata->sections;
+	$description = trim( $sections['description'] ) ?? '';
```
costdev: Need to trim the result of `$sections['description'] ?? ''` instead.

Trimming `$sections['description']` when it doesn't exist would throw a warning, and `trim()` returns a string, so `??` would never be hit.

```php
$description = trim( $sections['description'] ?? '' );
```

PR #288, opened 2025-10-07
Don't create a bundle for versions older than FAIR supports

johnbillion:
The bundles created for the release bundle FAIR with every version of WordPress back to 4.7, but FAIR only supports 5.4 and higher.

This adjusts the bundle script so it skips bundles for unsupported versions of WordPress, and fixes a path issue which was preventing `rsync` from running on my macOS machine.

## Steps to test:

1. Run `./bin/bundle.sh`
2. Confirm the output shows that older versions were skipped:

```
Skipping 5.3.20 (older than 5.4)
Skipping 5.2.23 (older than 5.4)
Skipping 5.1.21 (older than 5.4)
Skipping 5.0.24 (older than 5.4)
Skipping 4.9.28 (older than 5.4)
Skipping 4.8.27 (older than 5.4)
Skipping 4.7.31 (older than 5.4)
```

### DIFF
--- a/bin/bundle.sh
+++ b/bin/bundle.sh
@@ -22,7 +22,7 @@ touch /tmp/fair-dist/SHA384SUMS
 # Bundle our plugin first.
 [ -d /tmp/fair-temp ] && rm -rf /tmp/fair-temp
 mkdir -p /tmp/fair-temp/wordpress/wp-content/plugins/fair-plugin
-rsync -a --exclude-from="$SCRIPT_DIR/../.distignore" "$SCRIPT_DIR/.." /tmp/fair-temp/wordpress/wp-content/plugins/fair-plugin
+rsync -a --exclude-from="$SCRIPT_DIR/../.distignore" "$SCRIPT_DIR/../" /tmp/fair-temp/wordpress/wp-content/plugins/fair-plugin
 
 # Extract minimum required WordPress version from plugin header.
 REQUIRES_AT_LEAST=$(get_plugin_header "Requires at least")


kasparsd:
A question unrelated to this pull request -- what's the reason for creating all these releases? It definitely made me think before downloading the release for install. I'm wondering how users would know which version to pick? That information isn't part of the release notes or in the README.

fair-releases

p.s. happy to open an issue if we want to discuss that in isolation.

Ipstenu:
@kasparsd 

> I'm wondering how users would know which version to pick? 

Hopefully the users aren't using Github, but the answer is the one named `fair-plugin*`

The others (with `wordpress`) are installers for WP core with our plugin auto-added (like Akismet or Hello Dolly)

@rmccue I'm flagging this for the 1.2 release - I do think there's a point we don't need to build things we can't support :D

kasparsd:
> Hopefully the users aren't using Github

I believe the current instructions are to download the plugin from GitHub releases:

https://github.com/fairpm/fair-plugin/blob/4be6646767381670f0658a74c1e7a6d395fd325d/README.md?plain=1#L15-L17

https://github.com/fairpm/.github/blob/734cf38798dc9eeb8bdab1cba38ddbcd37bd48de/profile/README.md?plain=1#L9-L10

So I did that and was really overwhelmed with the amount of ZIPs I saw there 😅

All instances of "Install Plugin" appear to link to the GitHub releases:

- https://fair.pm/blog/2025/09/24/discover-trust-install-fair-1-0-is-here/
- https://fair.pm/packages/
- https://github.com/fairpm

Screenshots for reference:

tip

blog-post

PR #294, opened 2025-10-11
German translation for FAIR v. 1.0

La-Geek:
Updated German translation files for FAIR v. 1.0,
**included**: fair-de_DE.po, fair-de_DE.po.
**Missing** file: fair-de_DE.l10n.php


La-Geek:
Github is new for me, I have no idea, how to create/update: fair-de_DE.l10n.php
It would be great, if someone could create it 🙏

PR #302, opened 2025-10-27
Skip avatar URLs or link markup that do not contain `secure.gravatar.com`.

### LINKED ISSUE #142: Non-Gravatar profile images are replaced
I use [the WebMentions plugin](https://github.com/pfefferle/wordpress-webmention/) to get replies to my blog posts. User's images are taken from the account which replied.  FAIR Avatars overwrite those custom avatars as well as Gravatar.

For example - in the left screenshot, the first two images are Gravatars, the next two are from BlueSky, and the final one is a website's profile image. The right screenshot is what I see when FAIR avatars are enabled. I understand that the Gravatars will be replaced, but should the WebMention ones as well?

The code my theme uses to get comment avatars is:

```php
$avatar_html = get_avatar( $comment, 64, "", "", array( "extra_attr" => "itemprop=image loading=lazy" ) );
```

For a WebMention avatar, it returns something like:

```html
<img 
 alt=""
 class="avatar avatar-64 photo local-avatar"
 decoding="async"
 height="64"
 itemprop="image"
 loading="lazy"
 onerror="this.onerror=null;this.src='https://shkspr.mobi/blog/wp-content/plugins/webmention/assets/img/mm.jpg';this.srcset='https://shkspr.mobi/blog/wp-content/plugins/webmention/assets/img/mm.jpg';"
 src="https://cdn.bsky.app/img/avatar/plain/did:plc:fm7crvtisfj5xwzwuxerw32o/bafkreiepjfyygouxkrp4fybh7mqoo4b5kmk6c4wwsd52yn4hbs6ungewai@jpeg"
 srcset="https://cdn.bsky.app/img/avatar/plain/did:plc:fm7crvtisfj5xwzwuxerw32o/bafkreiepjfyygouxkrp4fybh7mqoo4b5kmk6c4wwsd52yn4hbs6ungewai@jpeg 2x"
 width="64">
```

For a Gravatar, it is the same, but the src is `src="https://shkspr.mobi/blog/wp-content/litespeed/avatar/f5eb5c3cbe6e929f18cbe191d4b95251.jpg?ver=1750834262"` because I'm using LiteSpeed to cache the images.

Would it be possible to check if an avatar is a Gravatar before replacing it?

It looks like the issue is in this code:

https://github.com/fairpm/fair-plugin/blob/52d2758b945ae9a57efb89add532dd0cb53623be/inc/avatars/namespace.php#L198

It is checking the author's ID, but perhaps it could also check if a comment has a set avatar as well?

Happy to provide more details if necessary.

costdev:
To ensure only Gravatar URLs are replaced, this PR checks for `secure.gravatar.com` in the avatar URL or link markup before making a replacement.

Fixes #142

### DIFF
--- a/inc/avatars/namespace.php
+++ b/inc/avatars/namespace.php
@@ -31,7 +31,6 @@ function bootstrap() {
 
 	// Filter avatar retrieval.
 	add_filter( 'get_avatar', __NAMESPACE__ . '\\filter_avatar', 10, 6 );
-	add_filter( 'get_avatar_url', fn () => 'https://upload.wikimedia.org/wikipedia/commons/b/b6/Image_created_with_a_mobile_phone.png', 10, 3 );
 	add_filter( 'get_avatar_url', __NAMESPACE__ . '\\filter_avatar_url', 10, 3 );
 	add_filter( 'avatar_defaults', '__return_empty_array' );
 


inc/avatars/namespace.php
```diff
@@ -143,6 +143,10 @@ function save_avatar_upload( $user_id ) {
  * @return string              Filtered avatar HTML.
  */
 function filter_avatar( $avatar, $id_or_email, $size, $default, $alt, $args ) {
+	if ( ! str_contains( $avatar, "src=''" ) && ! str_contains( $avatar, 'secure.gravatar.com' ) ) {
```
Ipstenu: ❓ Question - Should we make this a define or something? Just in case we have more to add later, and/or people want to be able to override?

Mostly thinking a single place to define `$avatar_domains_to_replace = [ 'secure.gravatar.com' ];`

costdev: Sounds good! Should be done in [5270536](https://github.com/fairpm/fair-plugin/pull/302/commits/5270536bae1a2c735b4bf29dd063b716638d1d12) - let me know if you're happy with that.

Ipstenu (approved):

afragen (approved):

PR #305, opened 2025-10-28
Modify Add Plugins message.

### LINKED ISSUE #41: Add Plugin screen - https://wordpress.org/plugins/ URL in the message below tabs
On the Add Plugin screen, below the tabs, there is a message 

>>>
Plugins extend and expand the functionality of WordPress. You may install plugins in the [WordPress Plugin Directory](https://wordpress.org/plugins/) right from here, or upload a plugin in .zip format by clicking the button at the top of this page.

Do we want to keep the https://wordpress.org/plugins/ URL or replace it with https://aspirepress.org/plugins/ ?

Image

afragen:
This replaces the current message on the Add Plugins page to refer to [FAIR Package Directory](https://fair.pm/packages/plugins)

Fixes #41 

Only issue is pre-6.9 the message for the **WordPress Plugin Directory** will continue to show unless we remove the **Featured** tab.

WP 6.8 and earlier
screenshot_733

WP 6.9 and later
screenshot_732

### DIFF
--- a/inc/packages/admin/namespace.php
+++ b/inc/packages/admin/namespace.php
@@ -40,7 +40,7 @@ function bootstrap() {
 	add_action( 'wp_ajax_check_plugin_dependencies', __NAMESPACE__ . '\\set_slug_to_hashed' );
 	add_filter( 'wp_list_table_class_name', __NAMESPACE__ . '\\maybe_override_list_table' );
 
-	// Needed for pre WP 6.9 compatibility.
+	// Needed for pre WordPress 6.9 compatibility.
 	add_action( 'install_plugins_featured', __NAMESPACE__ . '\\replace_featured_message', 5 );
 }
 
@@ -75,6 +75,7 @@ function add_direct_tab( $tabs ) {
 /**
  * Replace the featured message with our own.
  *
+ * @until WordPress 6.9.0
  * @return void
  */
 function replace_featured_message() {


inc/packages/admin/class-list-table.php
```diff
@@ -14,6 +14,40 @@
  * Custom plugin installer list table.
  */
 class List_Table extends WP_Plugin_Install_List_Table {
+
+	/**
+	 * Overrides parent views so we can use the filter bar display.
+	 */
+	public function views() {
+		$views = $this->get_views();
```
costdev: This works but does mean maintaining a copy of the whole parent method.

What about capturing the output of the parent method, then replacing the translated URL and "WordPress" so those are the only two things we're dependent upon?

```php
ob_start();
parent::views();
$views = ob_get_clean();

echo str_replace(
    // phpcs:ignore WordPress.WP.I18n.MissingArgDomain -- Intentional use of Core's text domain.
    [ __( 'https://wordpress.org/plugins/' ), __( 'WordPress' ) ],
    [ esc_url( 'https://fair.pm/packages/plugins/' ), 'FAIR' ],
    $views
);
```

costdev: (may need `wp_kses_post()` or similar for that `echo` though)

afragen:
Now works for our supported versions of WP

afragen:
screenshot_734

PR #309, opened 2025-10-30
Fix plugin search, broke with escaping

afragen:
Plugin search broke with escaping from #305

As replacement text is escaping we can ignore the other content as it's passed from core.


afragen:
Fixes #311

costdev:
Tested on nightly (6.9-beta2-61080-src) and 6.8, in both English (US) and Español

- Search works as expected in all combinations of the versions/locales above. ✅
- The plugin directory URL and text is replaced on a site using English (US), Español, and also when using Español and running a WordPress version that has no translation for the string (i.e. 6.9 beta has no translations yet, so it falls back to English). ✅

PR #328, opened 2025-11-12
fix for non-fair plugin install with no did string

### LINKED ISSUE #327: 🐞 Bug Report: Error when installing non-FAIR plugin
When installing a non-FAIR plugin there is no expanded did string attached to the slug. This results in an issue in `set_slug_to_hashed()` as there is a missing array key.

## Detailed Description
As above, a non-FAIR plugin installed via the Add Plugins page results in an error for a missing array key in `set_slug_to_hashed()`

```
Dependencies check failed: ( ! ) Warning: Undefined array key 1 in /Users/afragen/Documents/github/fair-plugin/inc/packages/admin/namespace.php on line 292
```

## Expected Behavior
No errors

## Current Behavior
Missing array key error when installing a dot org plugin.

## Steps to Reproduce

1. Go to Add Plugins
2. Try to install a dot org plugin
3. Error as above noted and Activate button doesn't load

## Possible Solution
See PR #328

afragen:
Add default to `explode()` in `set_slug_to_hashed()`
Fixes #327

### DIFF
--- a/inc/packages/admin/namespace.php
+++ b/inc/packages/admin/namespace.php
@@ -289,7 +289,7 @@ function set_slug_to_hashed() : void {
 	}
 
 	$escaped_slug = sanitize_text_field( wp_unslash( $_POST['slug'] ) );
-	$did = 'did:' . explode( '-did:', str_replace( '--', ':', $escaped_slug ), 2 )[1];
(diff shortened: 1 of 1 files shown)


afragen:
I don't think a doing it wrong is correct here. The issue is specifically about regular dot org plugin installation where a DID is not expected.

Ipstenu (approved):

PR #227, opened 2025-08-19
Hash the directory name of a package on install.

costdev:
This PR ensures that the directory name of a FAIR package is renamed to `slug-didhash` on installation.

In particular, this deals with the installation of a package without knowing its DID in advance. For example, when uploading the ZIP file of a plugin or theme.

PR #329, opened 2025-11-17
Maybe rename package on download

### LINKED ISSUE #228: Handle "Upload ZIP" and WP-CLI install location
### Overview
To prevent naming conflicts, we install each FAIR package to a `slug-didhash` directory, leaving `slug` for non-FAIR packages.

Currently, we handle this for packages installed via the "Direct Install" tab on the `Plugins > Add New Plugin` screen.

This issue is about handling packages installed via:
- `Plugins > Add New Plugin > Upload Plugin`
- `Appearance > Themes > Add New Theme > Upload Theme`
- `wp plugin|theme install <URL|path>`

Technically, implementation would most likely cover installation via the plugin and theme cards we see on the `Add New` screens too.

afragen:
Plugins are already renamed on download.

Slightly different take on #227 to fix #228

Use @costdev's `get_did_by_path()`

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -760,19 +760,18 @@ function delete_cached_did_for_install(): void {
  *
  * @return string|WP_Error
  */
-function maybe_rename_on_package_download( $source, string $remote_source, WP_Upgrader $upgrader, array $hook_extra ): string {
+function maybe_rename_on_package_download( $source, string $remote_source, WP_Upgrader $upgrader, array $hook_extra ) {
 	global $wp_filesystem;
 
-	// Early exit on passed errors.
+	$type = $upgrader instanceof Plugin_Upgrader ? 'plugin' : ( $upgrader instanceof Theme_Upgrader ? 'theme' : '' );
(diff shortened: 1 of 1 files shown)


inc/packages/namespace.php
```diff
@@ -703,8 +705,20 @@ function delete_cached_did_for_install(): void {
...
+			return $source;
+		}
+		$did_doc = get_did_by_path( $source, $type );
+		if ( is_wp_error( $did_doc ) ) {
+			return $did_doc;
+		}
+		$did = $did_doc->get_id();
+	}
+
 	if ( ! $did ) {
```
costdev: @afragen I'm not sure this condition can ever be true now. Either we bail on `empty( $type )` or `is_wp_error( $did_doc )`, but otherwise, `$did === $did_doc->get_id()`.

Any idea if there's some way that this condition can ever be true?

afragen: I don't think this will ever be true now

costdev: Removed in [24dee95](https://github.com/fairpm/fair-plugin/pull/329/commits/24dee95ace6d039c2a9775edb0c250e106841322).

Issue #257, opened 2025-09-22
Should we use `set|get_site_transient()` or `set|get_transient()`


afragen:
If the plugin is designed to be network activated on a multisite, is there a reason to use `set|get_transient()` as a subsite specific transient or should we use `set|get_site_transient()` so the transients work across all subsites of a multisite?

kasparsd:
I believe most of the functionality of this plugin runs at the network main site so unless there is a site-specific data (I can't think of any honestly), we can safely default to `*_site_transient()`. The opposite is also somewhat true -- since all of the update code runs only on the main network site, the site-level transients will never populate or get used. The overhead would only apply to code paths that run on every subsite.

Here is a report of all `_transient()` calls at revision 65be334f0ef03f407917de36e3fa960fe67049dc:

```
./inc/dashboard-widgets/namespace.php:92:       $response = get_transient( EVENTS_API );
./inc/dashboard-widgets/namespace.php:98:               set_transient( EVENTS_API, $response, CACHE_LIFETIME );
./inc/updater/class-lite.php:126:                       $response = get_site_transient( "git-updater-lite_{$this->file}" );
./inc/updater/class-lite.php:143:                               set_site_transient( "git-updater-lite_{$this->file}", $this->api_data, 5 * \MINUTE_IN_SECONDS );
./inc/updater/class-lite.php:260:               public function update_site_transient( $transient ) {
./inc/updater/class-lite.php:380:                       $current = get_site_transient( 'update_themes' );
./inc/updater/namespace.php:92: $did = get_transient( CACHE_DID_FOR_INSTALL );
./inc/updater/namespace.php:110:        $releases = get_transient( CACHE_RELEASE_PACKAGES ) ?? [];
./inc/updater/namespace.php:163:        $did = get_transient( CACHE_DID_FOR_INSTALL );
./inc/updater/class-updater.php:232:    public function update_site_transient( $transient ) {
./inc/updater/class-updater.php:321:            $current = get_site_transient( 'update_themes' );
./inc/packages/namespace.php:92:        $cached = get_transient( CACHE_METADATA_DOCUMENTS . $id );
./inc/packages/namespace.php:107:       set_transient( CACHE_METADATA_DOCUMENTS . $id, $document, CACHE_LIFETIME );
./inc/packages/namespace.php:152:       $response = get_transient( $cache_key );
./inc/packages/namespace.php:173:               set_transient( $cache_key, $response, CACHE_LIFETIME );
[...]

PR #330, opened 2025-11-17
Update get|set_transient to get|set_site_transient

afragen:
Fixes https://github.com/fairpm/fair-plugin/issues/257

### DIFF
--- a/inc/dashboard-widgets/namespace.php
+++ b/inc/dashboard-widgets/namespace.php
@@ -89,13 +89,13 @@ function get_community_events_ajax() : void {
  * @return array|WP_Error List of events or WP_Error on failure.
  */
 function get_community_events() {
-	$response = get_site_transient( EVENTS_API );
+	$response = get_transient( EVENTS_API );
 	if ( ! $response ) {
 		$response = wp_remote_get( EVENTS_API );
 		if ( is_wp_error( $response ) ) {
 			return $response;
 		}
-		set_site_transient( EVENTS_API, $response, CACHE_LIFETIME );
+		set_transient( EVENTS_API, $response, CACHE_LIFETIME );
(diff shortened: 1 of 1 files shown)


inc/dashboard-widgets/namespace.php
```diff
@@ -89,13 +89,13 @@ function get_community_events_ajax() : void {
  * @return array|WP_Error List of events or WP_Error on failure.
  */
 function get_community_events() {
-	$response = get_transient( EVENTS_API );
+	$response = get_site_transient( EVENTS_API );
```
norcross: this might be the one instance where we would want per-site, since different sites could have different locales?

afragen: Where does this pick up the locale from the constant?

inc/packages/namespace.php
```diff
@@ -703,7 +703,7 @@ function delete_cached_did_for_install(): void {
 function rename_source_selection( string $source, string $remote_source, WP_Upgrader $upgrader ) {
 	global $wp_filesystem;
 
-	$did = get_transient( CACHE_DID_FOR_INSTALL );
+	$did = get_site_transient( CACHE_DID_FOR_INSTALL );
```
costdev, line 706: Need to do the same for the function above when deleting the transient.

inc/dashboard-widgets/namespace.php
costdev: Yeah it seems like the locale won't be taken into consideration here. Is that likely to change in the future? If so, let's leave this as `set_transient()`, otherwise we can go ahead and change it to `set_site_transient()`.

afragen: OK

PR #332, opened 2025-11-17
Add array_first() and array_last() compatibility from WP6.9

afragen:
array_first() and array_last() compatibility functions added in WP 6.9


costdev (approved):
Matches [the implementations in WordPress 6.9](https://github.com/WordPress/wordpress-develop/blob/6.9/src/wp-includes/compat.php#L565-L605), although I'm not sure why/how the `$array` parameter name (and `// phpcs:ignore` comment) was committed to Core. 🤔

PR #337, opened 2025-11-23
rework dotorg_connectivity site health check - 2

joedolson:
This is a rebuild of [PR#30 ](https://github.com/fairpm/fair-plugin/pull/30) after I made a complete hash of rebasing following a missed signoff.

The work here is by @ideag, props due to @circuitaxel.

### DIFF
--- a/inc/site-health/namespace.php
+++ b/inc/site-health/namespace.php
@@ -55,7 +55,7 @@ function build_error_message_regex() {
 /**
  * Filter debug information.
  *
- * @param array $args {
+ * @param array $info {
  *     The debug information to be added to the core information page.
  *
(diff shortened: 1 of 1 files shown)


joedolson:
Now that this is my own PR, of course, I can't review it... I did, however, resolve the PHPCS issues from the previous PR.

joedolson:
Linking to related issue: #9

joedolson:
I've tested the JS filtering, and it's working to filter this message correctly when there's an error, but wasn't filtering the WordPress connectivity fields on the Info tab.

That's filterable in PHP using the `debug_information` filter, so I added a simple filter to handle that case.

PR #342, opened 2025-11-27
Appropriate slug rename on update

### LINKED ISSUE #341: 🐞 Bug Report: Hard set of package slug to `slug-didhash` if updated via FAIR Connect
## Detailed Description
If a package is installed prior to FAIR Connect being installed or the user wishes to install such that the slug is not modified via FAIR we should respect that decision. Currently if a package is updated via FAIR it is automatically modified to using the `slug-didhash` which can result in a deactivated plugin on update.

## Expected Behavior
It should just work as expected.

## Current Behavior
As above, a FAIR package updated via FAIR will have it's installed slug automatically changed to `slug-didhash` even if the package was installed as `slug`.

## Steps to Reproduce in local environment
1. Install FAIR Connect and activate.
2. Install Git Updater, decrease version number to force an update.
3. Update - this sets the DID transient
4. Decrease Git Updater version number to force an update.
6. Make sure installed slug is `git-updater/git-updater.php`
7. Make sure Git Updater is activated.
8. Run update and see if plugin still active.

## Context (Environment)
Updated plugin deactivated.

## Possible Solution
See associated PR

afragen:
Fixes a hard forced rename of a FAIR integrated package to the `slug-didhash`. This can be problematic for previously installed and running packages that suddenly are updating via FAIR and now have their install slugs changed and are automatically deactivated.

Fixes #341

### DIFF
--- a/inc/packages/admin/namespace.php
+++ b/inc/packages/admin/namespace.php
@@ -34,7 +34,7 @@ function bootstrap() {
 	add_action( 'install_plugins_' . TAB_DIRECT, __NAMESPACE__ . '\\render_tab_direct' );
 	add_action( 'load-plugin-install.php', __NAMESPACE__ . '\\load_plugin_install' );
 	add_action( 'install_plugins_pre_plugin-information', __NAMESPACE__ . '\\maybe_hijack_plugin_info', 0 );
-	add_filter( 'plugins_api_result', __NAMESPACE__ . '\\alter_slugs', 10, 3 );
+	add_filter( 'plugins_api_result', __NAMESPACE__ . '\\handle_did_in_search_results', 10, 3 );
 	add_filter( 'plugins_api_result', __NAMESPACE__ . '\\sort_sections_in_api', 15, 1 );
 	add_filter( 'plugin_install_action_links', __NAMESPACE__ . '\\maybe_hijack_plugin_install_button', 10, 2 );
 	add_filter( 'plugin_install_description', __NAMESPACE__ . '\\maybe_add_data_to_description', 10, 2 );
@@ -408,7 +408,7 @@ function maybe_hijack_legacy_plugin_info() {
  * @param object          $args   Plugin API arguments.
  * @return object|WP_Error
  */
-function alter_slugs( $res, $action, $args ) {
+function handle_did_in_search_results( $res, $action, $args ) {
 	if ( 'query_plugins' !== $action ) {
 		return $res;
 	}
@@ -462,7 +462,7 @@ function sort_sections_in_api( $res ) {
 /**
  * Override the install button, for bridged plugins.
  *
- * Bridged plugins appear in the API (via `alter_slugs()`) with slugs like
+ * Bridged plugins appear in the API (via `handle_did_in_search_results()`) with slugs like
  * `plugin-name-did--method--msid`, however they are installed to
  * `plugin-name-didhash`. In order to show the correct button, we need to check
  * against the install slug, not the API slug.


inc/packages/namespace.php
```diff
@@ -719,12 +719,12 @@ function rename_source_selection( string $source, string $remote_source, WP_Upgr
 		return $source;
 	}
 
-	if ( str_contains( $source, get_did_hash( $did ) ) && basename( $source ) === $metadata->slug ) {
+	if ( basename( $source ) === $metadata->slug ) {
 		return $source;
+	} else {
+		$new_source = trailingslashit( $remote_source ) . $metadata->slug . '-' . get_did_hash( $did );
```
costdev: I take it you added the `else` just to group the outcomes in the same block?

costdev:
Installing Git Updater by searching `Add Plugins` for "Git Updater" didn't rename the directory properly.

It ended up: `afragen-git-updater-cd3e9e9/git-updater.php`
On 1.1.0, the directory is correctly renamed: `git-updater-f27e06/git-updater.php`

Searching by DID appears to work fine though.

inc/packages/namespace.php
afragen: Yes

afragen:
Something more to test, thanks.

afragen:
Something more to test, thanks.

afragen:
> Installing Git Updater by searching `Add Plugins` for "Git Updater" didn't rename the directory properly.
> 
> It ended up: `afragen-git-updater-cd3e9e9/git-updater.php` On 1.1.0, the directory is correctly renamed: `git-updater-f27e06/git-updater.php`

This worked fine for me.

costdev:
> This worked fine for me.

Try it on the Playground link

afragen:
Sorry, but since I can't trace code in Playground, I have zero idea of what's happening and perhaps it's simply related to Playground.

afragen:
I see the issue. There is no DID associated with the Install button to later parse.

afragen:
Maybe AC can add the Plugin ID header to `$upgrader->new_plugin_data` and we could use that?

costdev:
> I see the issue. There is no DID associated with the Install button to later parse.

Yeah, noticing that myself too. That's not caused by this PR though. We're detecting and adding a DID to the release package cache whenever it's already installed on the site or when the user searches by the DID, but not if it appears in search results when they just search for a term.

I'll get a PR up for this separate issue ASAP.

For now, I'm going to try to work out why I was getting a different directory name outcome between this PR and 1.1.0. I'll either rule out this PR as the cause, or flag what's going on if in certain circumstances it causes a regression.

afragen:
It's because there's nothing from the DID cache and that's the standard zip from GitHub that we must rename. 

Since it's an install and not an update. We don't have the usual data. 

If the Plugin ID and Theme ID headers are passed in `$upgrader->new_plugin_data` we can get the DID and fix it.

afragen:
To test you will need to have `false` returned from the DID cache.

costdev:
Though `new_plugin_data` and `new_theme_data` aren't always set, depending on filesystem setups.

We just need to add any FAIR packages that appear in the search results to the release package cache, as we already do for direct DID searches and for updates. The DID is available in the search result data.

afragen:
Related https://github.com/fairpm/aspirecloud/issues/18

inc/packages/admin/namespace.php
```diff
@@ -418,14 +417,15 @@ function alter_slugs( $res, $action, $args ) {
...
-	// Alter the slugs to our globally unique version.
+	// Alter the slugs to our globally unique version and populate release cache.
 	foreach ( $res->plugins as &$plugin ) {
 		if ( ! is_fair_plugin( $plugin ) ) {
 			continue;
 		}
 
 		$did = $plugin['_fair']['id'];
 		$plugin['slug'] = esc_attr( $plugin['slug'] . '-' . str_replace( ':', '--', $did ) );
+		Packages\add_package_to_release_cache( $did );
```
costdev, line 428: While this is good for brevity, I think it looks out of place in an `alter_slugs()` function. Can we rename this function to `handle_did_in_search_results()` so it's clearer that it can encompass several actions?

costdev:
I've done some more testing since the recent updates and all paths I can think of seem to be working as intended now.

costdev (approved):

PR #347, opened 2025-12-04
removing data on uninstall

### LINKED ISSUE #144: add uninstall.php
overlooked during the initial scoping, there needs to be an uninstall function to clear our settings out

norcross:
this PR removes the site and network settings, along with the FAIR-unique user meta. fixes #144

### DIFF
--- a/uninstall.php
+++ b/uninstall.php
@@ -30,7 +30,7 @@ $query_args = $wpdb->prepare("
 		%s,
 		%s
 	)
-", esc_attr( 'fair_avatar_site_id' ), esc_attr( 'fair_avatar_id' ) );
+", 'fair_avatar_site_id', 'fair_avatar_id' );
 
 // And actually run it.
 // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared -- we did use the prepare function above


uninstall.php
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Include the avatar namespace so we can use the constant.
+require_once __DIR__ . '/inc/avatars/namespace.php';
+
+// Delete any single or multisite keys.
+delete_option( 'fair_indexnow_key' );
+delete_site_option( \FAIR\Avatars\AVATAR_SRC_SETTING_KEY );
+
+// Delete any transients we may have.
```
costdev:
```suggestion
// Delete any transients we may have or modify.
```

uninstall.php
```diff
@@ -0,0 +1,37 @@
...
+global $wpdb;
+
+// Prepare our delete query.
+$query_args = $wpdb->prepare("
+	DELETE FROM $wpdb->usermeta
+	WHERE       meta_key IN (
+		%s,
+		%s
+	)
+", esc_attr( 'fair_avatar_site_id' ), esc_attr( 'fair_avatar_id' ) );
```
afragen: Do we need to escape non-variable, known text strings? Did PHPCS report these?

costdev: PHPCS doesn't report these. They're hardcoded strings, and `wpdb->prepare()` leads to various escaping anyway. If the strings were coming from elsewhere, and used differently, then maybe `esc_sql()` would be used, but I don't think any additional escaping is necessary here.

costdev: Removed in [ee52e32](https://github.com/fairpm/fair-plugin/pull/347/commits/ee52e329c659e546abd1b30effd037ddf9ad5b0a).

costdev:
@afragen Any other thoughts for this one before we merge?

PR #363, opened 2025-12-11
Update conditional for possible empty $svg

afragen:
I found this issue when testing FAIR updating for FAIR.

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -520,7 +520,7 @@ function get_icons( $icons ) : array {
 
 	$icons_arr['1x'] = $regular->url ?? '';
 	$icons_arr['2x'] = $high_res->url ?? '';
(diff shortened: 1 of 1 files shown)


inc/packages/namespace.php
```diff
@@ -520,7 +520,7 @@ function get_icons( $icons ) : array {
 
 	$icons_arr['1x'] = $regular->url ?? '';
 	$icons_arr['2x'] = $high_res->url ?? '';
-	if ( str_contains( $svg->url, 's.w.org/plugins' ) ) {
+	if ( ! empty( $svg ) && str_contains( $svg->url, 's.w.org/plugins' ) ) {
```
```suggestion
	if ( ! empty( $svg->url ) && str_contains( $svg->url, 's.w.org/plugins' ) ) {
```
Might as well ensure the `url` isn't empty while we're here.

afragen:
Sounds fine

costdev (approved):

PR #364, opened 2025-12-11
Use get_transient as set_transient used in fair-beacon

afragen:
Somewhere along the way we strayed.

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -200,7 +200,7 @@ function fetch_metadata_from_local( $response, $url ) {
 	if ( ! $response && str_contains( $url, home_url() ) ) {
 		$did = explode( '/', parse_url( $url, PHP_URL_PATH ) );
 		$did = array_pop( $did );
-		$body = get_site_transient( 'fair-metadata-endpoint-' . $did );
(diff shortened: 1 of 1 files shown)


inc/packages/namespace.php
```diff
@@ -200,7 +200,7 @@ function fetch_metadata_from_local( $response, $url ) {
 	if ( ! $response && str_contains( $url, home_url() ) ) {
 		$did = explode( '/', parse_url( $url, PHP_URL_PATH ) );
 		$did = array_pop( $did );
-		$body = get_site_transient( 'fair-metadata-endpoint-' . $did );
+		$body = get_transient( 'fair-metadata-endpoint-' . $did );
```
costdev, line 203: I'm thinking it may be better to change Beacon to use `set_site_transient()` instead. 

While Beacon isn't a network-only plugin, Connect is, and I'm not sure there's a benefit/need to separating the metadata storage per site. Do you have any thoughts on that?

afragen: My mind is blank.

afragen:
Can close in favor of https://github.com/fairpm/fair-beacon/pull/84

PR #369, opened 2025-12-11
Fix jq command in workflow for closing PRs

rmccue:


rmccue:
Running this manually to verify.

Issue #367, opened 2025-12-11
🐞 Bug Report: php warnings "foreach() argument must be of type array|object, null given" in fair-plugin/inc/packages/namespace.php


futtta:
## Detailed Description

Since yesterday I have received 4 mails with PHP errors in cron output:

PHP Warning:  foreach() argument must be of type array|object, null given in /cust/0/futtta_14844/blogfu_25514/site/public/wp-content/plugins/fair-plugin/inc/packages/namespace.php on line 390

I also see those in Query Monitor on the backend

## Expected Behavior

No errors?

## Current Behavior

Errors?

## Context (Environment)

Running FAIR 1.1.0 on WordPress 6.9, hosting by Servebolt.

norcross:
thanks @futtta !

afragen:
I was chatting with @johnbillion in the past few days about this as he has added DIDs to User Switching and Query Monitor and is self-hosting using his toolset. I think some of the issues arose as we are behind in setting what protocol specs are required, especially for `ext-wp` and what are optional.

afragen:
Pretty sure this issue has been fixed. Please reopen if not.

Issue #350, opened 2025-12-04
🐞 Bug Report: Optional metadata properties aren't handled as such


johnbillion:
## Detailed Description

Various DID metadata properties are optional but lead to the FAIR plugin triggering PHP warnings when they're not present:

- `sections`
- ~~`filename` (not actually part of the FAIR spec)~~
- `release.requires`
- `release.suggests`

## Expected Behavior

Metadata properties that are optional should be handled as such.

## Current Behavior

PHP warnings are triggered in the following locations:

- https://github.com/fairpm/fair-plugin/blob/c47e9f2677bb976e8fd524f2710e5990d34ca529/inc/packages/namespace.php#L177
- https://github.com/fairpm/fair-plugin/blob/c47e9f2677bb976e8fd524f2710e5990d34ca529/inc/packages/namespace.php#L379
- https://github.com/fairpm/fair-plugin/blob/c47e9f2677bb976e8fd524f2710e5990d34ca529/inc/packages/namespace.php#L388
- https://github.com/fairpm/fair-plugin/blob/c47e9f2677bb976e8fd524f2710e5990d34ca529/inc/packages/namespace.php#L570

## Steps to Reproduce

1. Install a plugin which uses a FAIR metadata document that only contains the minimal fields that are required by the FAIR protocol spec.
   - If you're testing with existing plugin or data you'll need to flush transients
3. Trigger update checks.
4. Observe PHP errors from the FAIR plugin.

afragen:
https://github.com/fairpm/fair-protocol/pull/37 adds `filename` to spec

I think we are running into the issue of what is a requirement for the general spec and what is a requirement for WordPress. I believe the above are all above are WP-extensions and are required in that context.

norcross:
@johnbillion this was also reported in #367 and fixed in #372

PR #372, opened 2025-12-11
confirm array or object parts exist before using

norcross:
in addition to the error mentioned in #367 there is also a `PHP Warning:  Undefined property: stdClass::$sections in /wp-content/plugins/fair-plugin/inc/packages/namespace.php on line 177` error when a particular plugin does not have sections in their metadata (Query Monitor does this)

this PR addresses both of these instances


johnbillion:
I think this partially addresses #350

PR #378, opened 2025-12-14
Update rename on download

afragen:
add exit early if installing
update parameter type hinting
check for plugin and theme


afragen:
Oops, pretty sure I had that at some point in #329 but must've forgotten to move it.

afragen:
This will need to go into a 1.2.2 release.

PR #382, opened 2025-12-15
Add PHPStan checks

### LINKED ISSUE #376: 🚀 Feature Request: Add static analysis for PHP
## Describe the problem that needs this feature

Issues like #367 and #350 took down many sites because of errors in payload processing that could be caught and prevented by using static analysis. As the FAIR protocol gets adopted by more and more plugins and different repository hosts, it is important for the canonical FAIR user plugin to fail gracefully and account for various edge cases.

## Describe the solution you'd like

Introduce static analysis with `phpstan` that helps with catching errors like that.

## Describe alternatives you've considered

- Identify and document code paths that run on every WP request (including front-end and REST) and ensure those get special attention during testing and QA.

- Ensure that unit tests cover all happy-path scenarios and extend the testing suite to check edge cases as well.

kasparsd:
Fixes #376.

1. Add `szepeviktor/phpstan-wordpress` (includes both phpstan and the WP core stubs) at `v2.0` which supports the lowest supported PHP version 7.4 by the plugin.

2. Update the existing coding standard CI checks to run the PHPStan checks inside the same predictable `wp-env` environment. This is part of larger strategy of running all CI tasks using the same tooling as used for local development (`wp-env` in this case instead of a soup of bash scripts that rely on global mysql installs, etc.)

3. Update contributor documentation with all of the available helper scripts and their usage. Explain how to configure specific WP and PHP versions.


.github/workflows/coding-standards.yml
```diff
@@ -20,44 +19,40 @@ concurrency:
...
-        uses: actions/checkout@v4
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Setup cache
+        uses: actions/cache@v5
         with:
-          show-progress: ${{ runner.debug == '1' && 'true' || 'false' }}
+          key: ${{ runner.os }}-${{ hashFiles('composer.json') }} # Note that lock file will change between runs.
+          path: .cache
```
kasparsd, line 36: This is where both phpcs and phpstan store their caches.

.github/workflows/coding-standards.yml
```diff
@@ -20,44 +19,40 @@ concurrency:
...
 
-      - name: Install Composer dependencies
-        uses: ramsey/composer-install@3cf229dc2919194e9e36783941438d17239e8520 # v3.1.1
+      - name: Install dependencies
+        run: npm install
 
-      - name: Make Composer packages available globally
-        run: echo "${PWD}/vendor/bin" >> $GITHUB_PATH
+      - name: Start the development environment
+        run: npm run env start
```
kasparsd: Run the checks inside the same development environment as used locally. Ensures consistent results across local dev environments and the CI runs.

I've tested this with updated PHPUnit runs and they take the same ~1 min. as the runs currently.

.github/workflows/coding-standards.yml
```diff
@@ -20,44 +19,40 @@ concurrency:
...
+
+      - name: Install Composer dependencies
+        run: npm run test:php:install-deps
 
-      - name: Run PHPCS on all files
-        id: phpcs-files
-        run: phpcs . -n --report-full
+      - name: Lint PHPCS
+        # TODO: Remove the flag once all warnings are fixed.
+        run: npm run lint:php:phpcs -- -- -- --warning-severity=0
```
kasparsd: There are several existing warnings and errors reported (not sure why previously they didn't come up). I chose to not fix them as part of this pull request to make merging this easier.

```
FILE: inc/packages/admin/info.php
------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
------------------------------------------------------------------------------------------------
 498 | WARNING | Strings should not be wrapped in HTML (WordPress.WP.I18n.NoHtmlWrappedStrings)
------------------------------------------------------------------------------------------------


FILE: inc/packages/admin/namespace.php
------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
------------------------------------------------------------------------------------------------------------------------------------
 502 | WARNING | strip_tags() is discouraged. Use the more comprehensive wp_strip_all_tags() instead.
     |         | (WordPress.WP.AlternativeFunctions.strip_tags_strip_tags)
------------------------------------------------------------------------------------------------------------------------------------


FILE: inc/packages/namespace.php
------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
------------------------------------------------------------------------------------------------------------------------------------
 627 | WARNING | strip_tags() is discouraged. Use the more comprehensive wp_strip_all_tags() instead.
     |         | (WordPress.WP.AlternativeFunctions.strip_tags_strip_tags)
------------------------------------------------------------------------------------------------------------------------------------


FILE: inc/site-health/namespace.php
------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
------------------------------------------------------------------------------------------------------------------------------------
 1 | WARNING | A file should declare new symbols (classes, functions, constants, etc.) and cause no other side effects, or it
   |         | should execute logic with side effects, but should not do both. The first symbol is defined on line 13 and the first
   |         | side effect is on line 110. (PSR1.Files.SideEffects.FoundWithSymbols)
------------------------------------------------------------------------------------------------------------------------------------

Time: 241ms; Memory: 16MB
```

tests/phpstan-baseline.neon
```diff
@@ -0,0 +1,109 @@
+parameters:
```
kasparsd, line 1: To avoid making changes outside of adding the phpstan tooling to this pull request, we generate a baseline of "known issues" which can be fixed in follow-up pull requests.

composer.json
```diff
@@ -25,8 +26,19 @@
         }
     },
     "scripts": {
-		"lint": "@php ./vendor/bin/phpcs .",
-		"format": "@php ./vendor/bin/phpcbf .",
+        "lint:phpcs": "phpcs .",
+        "lint:phpstan": "phpstan analyse --verbose --memory-limit=2G",
+        "lint": [
```
kasparsd, line 33: We keep the existing `composer lint` as the combined task and introduce individual scripts for just `phpcs` and `phpstan` for convenience.

CONTRIBUTING.md
```diff
@@ -18,18 +18,29 @@ $ git commit -s -m 'My commit message.'
...
+By default wp-env is configured with PHP 7.4 (our minimum supported version), as well as Airplane Mode to avoid inadvertent requests.
 
-# Get the logs
-$ npm run env logs
+For linting and static analysis:
 
-# Stop the development server
-$ npm run env stop
-```
+- `npm run lint:php:phpcs` to run PHPCS (configured in [`phpcs.xml.dist`](phpcs.xml.dist)).
```
kasparsd, line 31: Documented a bunch of existing scripts along with pointers to their configuration.

Added instructions for overriding PHP and WP versions locally for running the tooling with any kind of setup. Also useful when debugging version-specific bugs and issues.

phpcs.xml.dist
```diff
@@ -47,6 +47,11 @@
 		<exclude-pattern>plugin\.php</exclude-pattern>
 	</rule>
 
+	<!-- Cache for PHPStan and PHPCS -->
+	<exclude-pattern>.cache</exclude-pattern>
```
kasparsd, line 50: These were being caught by the `phpcs`.

phpstan.dist.neon
```diff
@@ -0,0 +1,10 @@
+includes:
+	- tests/phpstan-baseline.neon # TODO: Fix all these issues.
```
kasparsd, line 2: This can be removed once we've fixed all the reported issues.

kasparsd:
@toderash Noticed this was moved to icebox. I'm not familiar with the triage/planning lingo for the project. Could you please confirm if there is any interest in merging this (or solving the associated #376) in near term? Or if there is anything required to get this merged?

I have a other follow-up pull requests ready that would fix #383, for example, and it relies on the cleanup and documentation updates in this pull request. Wanted to get a sense of when and how that could happen, if at all. Thanks!

kasparsd:
Noting that #394 would have been caught by phpstan if I'm not mistaken.

.github/workflows/coding-standards.yml
chuckadams: I'm rather 👎 on wp-env in general: it's running Node to run Docker to run PHP, when PHP directly will do fine in a controlled environment like a github runner.  This is a wordpress plugin, so PHP is going to be a dev requirement regardless, and phpstan's results are going to be identical across OS distributions.

It looks like we're already using wp-env anyway, so if there's no impact on runner time, I don't have a problem either way.  It just seems far more convoluted than it has to be.

.github/workflows/coding-standards.yml
```diff
@@ -20,44 +19,40 @@ concurrency:
...
```
chuckadams: run: npm run lint:php:phpcs -- -- -- --warning-severity=0

 Are there really supposed to be three occurrences of `--` here?  ~~Mainly though, if we have a baseline file, we shouldn't need to bring down the severity, no?~~  (edit: nvm I see this was for phpcs, not phpstan)

chuckadams:
I'm all for phpstan, but I refuse to deal with wp-env anymore:

```
❯ npm run lint:php:phpcs

> lint:php:phpcs
> wp-env run cli --env-cwd=wp-content/plugins/plugin composer run lint:phpcs

ℹ Starting 'composer run lint:phpcs' on the cli container. 

open /Users/chuck/.wp-env/ea95b44cf8f0dbec465691910de921ef/docker-compose.yml: no such file or directory
✖ Command failed with exit code 1
Command failed with exit code 1
```

`composer run lint:phpstan` does work fine though.

chuckadams:
In the interests of getting this merged before 1.4 release (even though it's a dev feature and the release doesn't matter) I'm going to patch this PR up myself so that the GH action does not use wp-env, then we can merge the rest to add the phpstan checks.  While an individual CI run might not have added much overhead, we have a large test matrix that magnifies the extra overhead of Node+Docker significantly.

chuckadams (dismissed):
Updated coding-standards.yml to more directly invoke the linting CI scripts, but leaving the wp-env stuff as-is in CONTRIBUTING.md, assuming it works for others.  Approved the workflow and it passes now.

cdils (dismissed):

cdils (approved):

PR #387, opened 2025-12-18
Update ca POT translation (12-18-25)

marcarmengou:


afragen:
@marcarmengou thanks for the translation. We
Only require the .po file as there is a GitHub Action to generate the other files.

marcarmengou:
Thanks @afragen, I'll keep that in mind for next time.

PR #422, opened 2026-01-13
Add FAIR publishing workflow and scripts

namithj:
Introduces FAIR-Pulse publishing support with GitHub Actions workflows and supporting PHP scripts for key generation, DID creation, artifact signing, and metadata generation. Includes detailed setup instructions in README


namithj:
The files are copy pasted from here : https://github.com/fairpm/fair-pulse (https://github.com/fairpm/fair-pulse/releases/tag/0.0.1) and uses the DID Manager Library https://github.com/fairpm/did-manager

PR #428, opened 2026-01-29
Rearchitect `Updater` into a registry

### LINKED ISSUE #427: 🐞 Bug Report: Every FAIR-enabled plugin or theme results in a new `Updater` instance adding its hooks
A new instance of the `Updater` class gets instantiated for every FAIR-enabled plugin or theme in place on a site. This causes all of the hooks in the `load_hooks()` method to be added and executed once per plugin and theme, instead of once globally.

## Detailed Description

This causes hooks such as `upgrader_source_selection`, `plugins_api`, `site_transient_update_plugins`, `wp_prepare_themes_for_js`, and `upgrader_pre_download` to be run `n` times instead of once.

This causes a few effects, including:

- `Updater::upgrader_source_selection()` attempts to rename and move every plugin and theme `n` times instead of once.
- `verify_signature_on_download()` downloads and verifies every package `n` times instead of once. As far as I can tell, there is no poisoning of the trusted key logic between plugins but I think this is by luck rather than by design.
- `Updater::repo_api_details()` runs `n` times instead of once, although this does contain a guard condition so it doesn't re-fetch package data `n` times, although it does parse and recalculate the DID hash `n` times instead of once.
- `Updater::customize_theme_update_html()` runs `n` times instead of once, which calls `Updater::append_theme_actions_content()` `n` times instead of once. I don't have any FAIR-enabled themes installed, but if I had more than one I anticipate that this would cause duplicate update messages to appear for a theme with a pending update.

The other effect this has is that I am unable to propose some enhancements that I want to make to `verify_signature_on_download()` because it unexpectedly runs with the incorrect DID context `n-1` times out of `n`.

## Expected Behavior

The hooks that the `Updater` class hooks into and handles should only be instantiated once. I think this means that the `Updater` class needs to be rearchitected so it acts as a registry and dispatches any plugin- or theme-specific logic as necessary.

johnbillion:
Fixes #427

I've been using this on my local installation for a couple of weeks, although with a fairly limited number of updates. Needs some thorough testing by the FAIR Connect team.

## Testing

See my comments on #427 about installing the add-on plugin for Query Monitor to see the duplicate hooks.

Testing this requires installing more than one plugin or theme that has a `Plugin ID: did:plc:...` header. You can install Pods, Query Monitor, User Switching, and WP Crontrol for testing this. They don't need to be active.

Prior to testing you can choose to flush all your transients or not. This change doesn't affect the way transients for updates are stored or handled so it will continue to work with your existing cache.

In order to trigger and test an update check for a plugin or theme, you'll need to clear the appropriate transients.

Items to test:

- Update checks
- Update notices
- Performing updates

### DIFF
--- a/inc/updater/class-updater.php
+++ b/inc/updater/class-updater.php
@@ -131,18 +131,20 @@ class Updater {
 
 		// Rename plugins.
 		if ( $upgrader instanceof Plugin_Upgrader ) {
-			if ( isset( $hook_extra['plugin'] ) ) {
-				$slug       = dirname( $hook_extra['plugin'] );
-				$new_source = trailingslashit( $remote_source ) . $slug;
+			if ( ! isset( $hook_extra['plugin'] ) ) {
+				return $source;
 			}
+			$slug       = dirname( $hook_extra['plugin'] );
+			$new_source = trailingslashit( $remote_source ) . $slug;
 		}
 
 		// Rename themes.
 		if ( $upgrader instanceof Theme_Upgrader ) {
-			if ( isset( $hook_extra['theme'] ) ) {
-				$slug       = $hook_extra['theme'];
-				$new_source = trailingslashit( $remote_source ) . $slug;
+			if ( ! isset( $hook_extra['theme'] ) ) {
+				return $source;
 			}
+			$slug       = $hook_extra['theme'];
+			$new_source = trailingslashit( $remote_source ) . $slug;
 		}
 
 		if ( basename( $source ) === $slug ) {


afragen:
@johnbillion how do I setup QM to see the hooks you are seeing?

inc/updater/class-updater.php
```diff
@@ -158,7 +109,7 @@ public function load_hooks() {
 	 *
 	 * @return string|WP_Error
 	 */
-	public function upgrader_source_selection( string $source, string $remote_source, WP_Upgrader $upgrader, $hook_extra = null ) {
+	public static function upgrader_source_selection( string $source, string $remote_source, WP_Upgrader $upgrader, $hook_extra = null ) {
```
afragen: Shouldn't this have the changes from `string $source` to `$source` and the early exit for a WP_Error?

johnbillion:
@afragen Can you change the target branch of this PR to `release_1.3.0` please? I don't know why I don't have permission to do it.

Ipstenu:
@afragen @chuckadams I forced DCO to pass for your commits. Please remember to sign off :)

inc/updater/class-updater.php
afragen:
```suggestion
	public static function upgrader_source_selection( $source, string $remote_source, WP_Upgrader $upgrader, $hook_extra = null ) {
```

Can't call non-static method statically 😉

afragen:
@cdils any reason this was closed?

cdils:
@afragen This should not have been closed. Looks like there's an auto-deletion on the `release_1.3.0` branch after the merge and that triggered closing open PRs against that branch. Clearly some kinks to work out. I'll flag this in Slack.

johnbillion:
This PR has also been moved into the "Done" column in the three projects. And it's been moved into the 1.4.0 milestone. I don't have permission to update either of those and I can't reopen the PR, so I'll just create a new one.

PR #429, opened 2026-02-07
Update browser check

### LINKED ISSUE #239: Recent iOS devices show "You are using an insecure browser!"
Recently-updated iOS devices (iPhones/iPads) now incorrectly trigger the BrowseHappy "You are using an insecure browser" warning. This appears to be addressed in the already-merged https://github.com/fairpm/fair-plugin/pull/214, but there hasn't been a release in the last several weeks. Thus, anyone keeping their devices up-to-date is seeing a warning that paradoxically claims their device is out-of-date.

A new release will also confirm that the version-number update code added in the last update works 🤣

joedolson:
Draft to fix #239

This PR adds the code used in the browse happy API, with some modifications for our  needs. 

We could add the images, but I don't feel that they're very necessary.

### DIFF
--- a/inc/version-check/namespace.php
+++ b/inc/version-check/namespace.php
@@ -567,7 +567,7 @@ function parse_user_agent( $user_agent ) {
 		} else {
 			$data['name'] = 'unknown';
 		}
-	} elseif ( $found = array_intersect( array_keys( $explicit_tokens ), array_keys( $tokens ) ) ) {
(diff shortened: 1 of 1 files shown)


Ipstenu:
Approving Andy's commits with his consent.

kasparsd:
Should we add attribution to WP core if we're forking those helpers? Not sure what the right approach is here.

joedolson:
@kasparsd It does make sense; need to figure out exactly what we're attributing to, though. This isn't part of core, so it doesn't make sense to attribute it there. There are only 17 contributors to the repository, however, and it probably wouldn't be difficult to figure out which people specifically to credit for this code.

https://github.com/WordPress/browsehappy

PR #461, opened 2026-03-10
Update stable tag version to 1.4.0

marcarmengou:
The stable version in readme.txt has been updated to the current version of the plugin.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -6,7 +6,7 @@ Tags: packages, updater, installer, technical independence
 Requires at least: 5.4
 Requires PHP: 7.4
 Tested up to: 6.9
-Stable tag: 1.1.0
+Stable tag: 1.3.0
 
 FAIR is a system for using **F**ederated **a**nd **I**ndependent **R**epositories in WordPress.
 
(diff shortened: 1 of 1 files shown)


readme.txt
```diff
@@ -6,7 +6,7 @@ Tags: packages, updater, installer, technical independence
 Requires at least: 5.4
 Requires PHP: 7.4
 Tested up to: 6.9
-Stable tag: 1.1.0
+Stable tag: 1.3.0
```
Ipstenu: Can we change this to 1.4.0 and merge this into the next release?

It's supposed to be automated but seems to be acting up...

Ipstenu (changes requested):
Can we change this for 1.4.0 instead? 1.3.0 would need a release to fix it, and we're working on 1.4.0 which will let us handle this.

readme.txt
marcarmengou: Yes, definitely. I probably should have just created an issue. Thanks!

marcarmengou (commented):
Update Stable tag to 1.4.0

cdils (approved):

cdils (approved):

PR #464, opened 2026-03-21
Add parser config.

zzap:
This is the config for the Parser created at CloudFest hackathon 2026 https://hackathon.cloudfest.com/project/block2docs/


zzap:
Too early

PR #465, opened 2026-03-28
Did manager integration

namithj:
Switch to DID Manager


plugin.php
```diff
@@ -24,6 +24,7 @@
 const PLUGIN_DIR = __DIR__;
 const PLUGIN_FILE = __FILE__;
 
+require_once __DIR__ . '/vendor/autoload.php';
```
afragen, line 27: There's no vendor/ committed so we'll need to either include it or...

namithj: It will come from composer install.

afragen: And where does composer install happen? I don't see it in the releases.yml.

namithj: Will add it.  Since the composer.json was already there, I expected the release action would already have the process.

namithj: Done via 25fd338816b56a2f452fc436a2522bf068e6596e

chuckadams:
Just reminding myself: add a reusable (copy-paste) check in the plugin load process to check for vendor/ or existing autoloader, and render a friendly hint in `wp_error` to run composer install.

cdils (approved):

PR #466, opened 2026-04-06
Fix most errors from PHPCS and PHPStan

chuckadams:
Not all issues are fixed yet, but I've left warnings enabled and the baseline disabled so they can keep failing in CI til we fix them here.  The FIXME comment in inc/packages/admin/namespace.php suggests there's probably a genuine bug in that code path.


inc/icons/svg.php
```diff
@@ -24,6 +24,8 @@ function sanitize_hex_color( $color ) {
 	if ( preg_match( '|^#([A-Fa-f0-9]{3}){1,2}$|', $color ) ) {
 		return $color;
 	}
+
+	return null;
```
afragen, line 28: I'm not sure I understand this? Why return null? Why not just return '' as in line 20?

inc/packages/admin/namespace.php
```diff
@@ -124,6 +124,7 @@ function handle_did_during_ajax( $result, $action, $args ) {
 		return $result;
 	}
 
+	// FIXME: Updater\Updater has neither a constructor nor a run() method.
```
afragen, line 127: Well, they both existed in 1.3.0

inc/icons/svg.php
chuckadams: I'm just replicating the behavior of the WP function which falls off the end, thus implicitly returning null.  phpstan is just forcing us to be explicit about that.  Frankly I'm not sure why it's not just using the WP function, but it's only used in one place anyway, so 🤷

inc/packages/admin/namespace.php
chuckadams: `FAIR\Updater\Updater` has nothing but static functions, so maybe it's just using the wrong class?  `Fair\Updater\Package` does take a $did argument (but also requires a second arg of $filepath).  I'm not familiar enough with this part of the code to know what's right, though my IDE does mark the surrounding function as unused...

chuckadams:
I've now fixed or suppressed the remaining warnings and errors, except for the one phpstan error that's a legit bug, and that needs someone who knows the code better.

inc/icons/svg.php
afragen: Why not just return an empty string if it falls off instead of null?

kasparsd:
Noting that all of the phpstan issues are being fixed in #517 too. I recommend we tackle this in parts and avoid mixing changes to CI, release logic and meta files (like `composer.json` formatting) in one changeset. This can serve as a nice reference for those individual cleanups, though!

PR #469, opened 2026-04-09
Release: merge 1.4.0 into development for BETA testing

cdils:


chuckadams:
I suggest `git rebase development` to update this branch, which leaves only two conflicting files, one of which is trivial.  The other one is `class-updater.php`, which I really don't know how to resolve, since this branch creates a large diff that appears to be mostly reverted in `development` and I don't know the whole story behind the changes.  

I suspect it's a matter of choosing one or the other, but since the API is changed incompatibly, use sites may have to be updated either way

Ipstenu (approved):

PR #471, opened 2026-04-15
add error checking in add_package_to_release_cache()

afragen:
Add error checking to `Packages::add_package_to_release_cache()`.

Don't save a WP_Error to the cache. `Packages::get_latest_release_from_did()` can return a WP_Error. We need to account for that.

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -938,6 +938,9 @@ function add_package_to_release_cache( string $did ) : void {
 	}
 	$releases = get_site_transient( CACHE_RELEASE_PACKAGES ) ?: [];
 	$releases[ $did ] = get_latest_release_from_did( $did );
+	if ( is_wp_error( $releases[ $did ] ) ) {
+		unset( $releases[ $did ] );
+	}
 	set_site_transient( CACHE_RELEASE_PACKAGES, $releases );
 }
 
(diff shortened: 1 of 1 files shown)


inc/packages/namespace.php
```diff
@@ -938,6 +938,9 @@ function add_package_to_release_cache( string $did ) : void {
 	}
 	$releases = get_site_transient( CACHE_RELEASE_PACKAGES ) ?: [];
 	$releases[ $did ] = get_latest_release_from_did( $did );
+	if ( is_wp_error( $releases[ $did ] ) ) {
+		unset( $releases[ $did ] );
+	}
 	set_site_transient( CACHE_RELEASE_PACKAGES, $releases );
```
```suggestion
 	$latest_release = get_latest_release_from_did( $did );
	if ( is_wp_error( $latest_release ) ) {
		return;
	}
 	$releases = get_site_transient( CACHE_RELEASE_PACKAGES ) ?: [];
 	$releases[ $did ] = $latest_release;
 	set_site_transient( CACHE_RELEASE_PACKAGES, $releases );
```

While I understand the approach you were using, I think this makes the method read a bit better overall in light of the error checking.

Additionally, it means there's no call to the transient API, which could be a database call rather than external object cache, unless necessary.

costdev (approved):

PR #493, opened 2026-05-15
Consistent phpunit runners for local and CI

### LINKED ISSUE #383: 🚀 Feature Request: Use wp-env tooling for PHPUnit CI runs
## Describe the problem that needs this feature

The current CI configuration for PHPUnit tests relies on globally installed versions of phpunit and mysql server which complicates the setup and prevents users from running the test suite in the same way locally.

The project has disabled `composer.lock` to allow running different versions of dependencies with various PHP/WP version combination. This leads to unexpected dependencies being installed in local development environments and also during other CI runs (outside of phpunit).

## Describe the solution you'd like

Re-use the same wp-env tooling for running phpunit with a matrix of PHP and WP core version combinations. It takes the same amount of time per run (~1 min.) and ensures that the developer experience is the same between local development and CI runs. 

Instead of disabling the `composer.lock` file for the project, we can simply remove it during test runs to ensure that locally users always get the same development tools and can reliably write unit tests and debug issues.

### LINKED ISSUE #479: 🚀 Feature Request: Review and update Unit tests
Check Unit tests to ensure they're properly running. Additionally, since Connect now has a minimum PHP v8.0 requirement, tests should reflect this.

https://github.com/fairpm/fair-plugin/actions/workflows/phpunit-tests.yml

kasparsd:
Fixes #479, fixes #383.

A follow-up to #382.

Development workflow consistency:

- Standardizes local development and CI around `wp-env` so both paths use the same environment bootstrapping and runtime logic.
- Adds a small npm command layer for common tasks like CLI access, linting, formatting, PHPUnit, and coverage, so developers and CI invoke the same commands.
- Updates the coding standards workflow to run through the `wp-env`-based toolchain instead of relying on a separate runner-level PHP setup.

Local setup and onboarding:

- Commits a Node version via `.nvmrc` and upgrades `@wordpress/env`, reducing local setup drift and keeping machines closer to CI.
- Expands the contributing docs to make `wp-env` the default local workflow, covering start/log/stop commands, PHPUnit usage, coverage, and version overrides.
- Supports both `.wp-env.override.json` and `WP_ENV_PHP_VERSION` / `WP_ENV_CORE` env vars, making local environment customization straightforward without changing repo defaults.

Local PHPUnit execution:

- Moves PHPUnit execution into the `wp-env` tests container, removing the old split between custom local setup and CI-only test installation.
- Adds a dedicated `tests:cli` path and install flow for test dependencies, so local test runs are easier to reason about and reuse.
- Aligns the single-site and multisite PHPUnit config files with this workflow, including cleaner cache and coverage paths.

WordPress and PHP version matrix support:

- Makes it possible to run PHPUnit locally against different WordPress core and PHP combinations by driving `wp-env` from override files or environment variables.
- Broadens Composer dev tooling to support multiple PHPUnit versions and include the WP-CLI bundle needed for this workflow.
- Replaces the hardcoded GitHub Actions exclusion list with a generated PHP/WordPress compatibility matrix, centralizing supported combinations in one script.

CI parity with local testing:

- Uses the generated matrix in GitHub Actions so each job boots the requested WordPress and PHP combination the same way it can be done locally.
- Refreshes Composer dependencies per matrix job so the right PHPUnit version is installed for that exact WordPress/PHP pairing.
- Extends the same `wp-env`-based approach to coverage runs, improving parity between local verification and CI.

## CI timing

Previously the `phpunit` setup in CI didn't use `wp-env` leading to two different ways of running the same tests. Switching to `wp-env` ensures that we can run the tests with any combination of PHP and WP core both locally and during CI.

Running the `phpunit` test matrix in CI before the change took 1min 42s:

matrix-before

while now it takes 3min 16s (or 2min 30s in recent runs):
 
matrix-after

which is a minor increase but offers consistency and predictability.


.github/workflows/coding-standards.yml
```diff
@@ -35,11 +35,25 @@ jobs:
           key: ${{ runner.os }}-${{ hashFiles('composer.json') }} # Note that lock file will change between runs.
           path: .cache
 
-      - name: Set up PHP
-        uses: shivammathur/setup-php@cf4cade2721270509d5b1c766ab3549210a39a2a # v2.33.0
+      - name: Setup Node
+        uses: actions/setup-node@v6
+        with:
+          cache: 'npm'
+          node-version-file: '.nvmrc'
```
kasparsd, line 45: Introduce `.nvmrc` to lock the version of Node to a known version for this project.

.github/workflows/phpunit-tests.yml
```diff
@@ -16,100 +16,119 @@ concurrency:
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
+      matrix: ${{ steps.generate.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v6
+      - name: Generate PHP and WP testing matrix
+        id: generate
+        run: echo "matrix=$(node ./bin/github-actions-matrix.js)" >> "$GITHUB_OUTPUT"
```
kasparsd, line 33: Use a simple node script to generate the test matrix since it is easier to read and maintain.

.github/workflows/phpunit-tests.yml
```diff
@@ -16,100 +16,119 @@ concurrency:
...
-          echo "define('WP_TESTS_PHPUNIT_POLYFILLS_PATH', '$HOME/.composer/vendor/yoast/phpunit-polyfills');" >> ${{ runner.temp }}/wordpress-tests-lib/wp-tests-config.php
-          phpunit
+          npm run env run cli -- php -v
+          npm run env run cli -- php -i | grep xdebug
+          npm run env run cli -- wp core version
+
+      - name: Install Composer dependencies
+        run: npm run test:php:install-deps
+
+      - name: Generate coverage report
```
kasparsd, line 131: Add in coverage reporting.

inc/icons/svg.php
```diff
@@ -15,15 +15,13 @@
  *
  * @return string
  */
-function sanitize_hex_color( $color ) {
-	if ( '' === $color ) {
-		return '';
-	}
-
+function sanitize_hex_color( $color ): string {
```
kasparsd, line 18: Fix a few remaining `phpstan` errors.

composer.json
```diff
@@ -15,18 +15,20 @@
         "fairpm/did-manager-wordpress": "0.0.4-a"
     },
     "require-dev": {
-        "yoast/phpunit-polyfills": "*",
-        "nimut/phpunit-merger": "*",
+        "yoast/phpunit-polyfills": "^4.0",
```
kasparsd, line 18: Lock everything to the versions supported by the min-supported PHP of the project (8.0). During CI we delete the lock file and update the requirements to work with the relevant versions of PHP.

composer.json
```diff
@@ -41,33 +43,25 @@
         "phpstan-baseline": "phpstan analyse --generate-baseline=tests/phpstan-baseline.neon --memory-limit=2G",
         "test": [
             "Composer\\Config::disableProcessTimeout",
-            "@php ./vendor/phpunit/phpunit/phpunit"
+            "phpunit"
         ],
         "test:multisite": [
             "Composer\\Config::disableProcessTimeout",
-            "@php ./vendor/phpunit/phpunit/phpunit -c tests/phpunit/multisite.xml"
```
kasparsd, line 48: Composer is smart enough to try the local `vendor/bin` binaries before using globals so we make this consistent with how `phpcs` and `phpstan` is called in this file.

phpcs.xml.dist
```diff
@@ -33,17 +33,15 @@
 	</rule>
 
 	<!-- Exclude third-party dependencies -->
-	<exclude-pattern>inc/updater/class-lite\.php</exclude-pattern>
+	<exclude-pattern>inc/updater/class-lite.php</exclude-pattern>
+	<exclude-pattern>inc/updater/class-base58btc.php</exclude-pattern>
```
kasparsd, line 37: Add the missing excludes and skip escaping `.` since it isn't necessary.

phpcs.xml.dist
```diff
@@ -53,34 +51,34 @@
 
 	<!-- Exclude the unit tests from the file name rules -->
 	<rule ref="WordPress.Files.FileName">
-		<exclude-pattern>tests/phpunit/*</exclude-pattern>
```
kasparsd, line 56: The trailing `/*` is also not necessary -- it matches by fragment by default.

phpstan.dist.neon
```diff
@@ -4,6 +4,8 @@ includes:
 parameters:
 	tmpDir: .cache/phpstan
 	level: 0
+	scanDirectories:
```
kasparsd, line 7: Ensure `phpstan` is aware of the wp-cli signatures.

phpunit-multisite.xml.dist
```diff
@@ -9,30 +9,31 @@
...
+			<exclude>tests/phpunit/tests/SampleTest.php</exclude>
 		</testsuite>
 	</testsuites>
 	<groups>
 		<exclude>
 			<group>ms-excluded</group>
 		</exclude>
 	</groups>
-	<coverage includeUncoveredFiles="true" processUncoveredFiles="false" pathCoverage="false" cacheDirectory="./cache">
+	<coverage includeUncoveredFiles="true" processUncoveredFiles="false" pathCoverage="false" cacheDirectory="tests/phpunit/cache">
```
kasparsd, line 29: Make this consistent with what `phpunit.xml.dist` was using to cache the results.

chuckadams:
I personally have less than zero interest in using wp-env as the standard environment: a dependency on node in order to run bash scripts in order to run PHP.   If this is the easiest way to switch WP versions quickly in CI, especially if it can cache the different versions, then I'm not going to stand in the way, but I would very much prefer to avoid gratuitous Node dependencies as far as the officially sanctioned dev workflow goes, and would rather see something like DDEV be it instead.

kasparsd:
> if this is the easiest way to switch WP versions quickly in CI, especially if it can cache the different versions

Yeah, this setup (and the updated documentation) allows us to consistently test the plugin against any version of PHP and WP core. Previously the setup used locally (`wp-env`) was different from what the CI would use, so there was no clear way to replicate failing builds.

I absolutely don't love `wp-env` but it gets the job done and is a shared convention that others know how to debug and use. The previous mix of bash scripts, system dependencies installed only in CI to run the same tests and linters was arguably more convoluted and harder to maintain.

cdils:
@kasparsd Beyond resolving conflicts, are you waiting for further review?

kasparsd:
@cdils if we agree on the concept (same tooling locally and in CI for consistency) then I believe this is ready for merge.

I'll resolve the merge conflicts now.

kasparsd:
@cdils the conflicts have been resolved ✅

Ipstenu (approved):

PR #496, opened 2026-05-16
Account for package metadata that doesn't include filename when deriving hashed filename

### LINKED ISSUE #492: 🐞 Bug Report: Add plugins "unexpected error" (FAIR 1.4.0)
After installing FAIR 1.4.0, an unexpected error occurs when going to the "add plugin" page. I note that this did not work with the previous version either, but that it did work for me about 6 months ago using a slightly older version of wordpress and FAIR (sorry, I forget which versions). Related to this, it is not possible to upgrade any plugins when FAIR is activated.

Image

## Detailed Description
FAIR: 1.4.0
Wordpress: 6.9.4
PHP: 8.3.29

kasparsd:
Fixes #492.

Updated [`get_hashed_filename()` method](https://github.com/fairpm/fair-plugin/blob/707135cbb5243c7cd83250c08ae93961630bf86e/inc/packages/namespace.php#L682-L705) to build a stable, hashed install path for a FAIR package from its metadata.

It works in this order:

1. It derives a short DID hash from the package ID.
2. It tries to split `filename` into two parts: a slug-like directory name and a plugin file name.
3. If the filename does not provide a slug, it falls back to the metadata slug.
4. If the filename does not provide a file portion, it assumes the main plugin file should be `slug.php`.
5. It appends the DID hash to the slug unless it is already present.
6. If the package type is `wp-plugin`, it returns a WordPress-style plugin path in the form `slug-didhash/file.php`.
7. For non-plugin types, it returns only the hashed slug.

The intent of this version is:

- prefer the slug embedded in `filename` when available
- gracefully recover when `filename` is missing or incomplete
- avoid double-appending the DID hash
- return plugin paths as `directory/file.php`
- return non-plugin paths as just the hashed directory/slug name

Examples for this implementation:

- plugin with `filename = example/example.php` returns `example-<hash>/example.php`
- plugin with missing filename falls back to `example-<hash>/example.php`
- theme with slug `example` returns `example-<hash>`

### DIFF
--- a/inc/default-repo/namespace.php
+++ b/inc/default-repo/namespace.php
@@ -26,7 +26,7 @@ function bootstrap() {
  */
 function get_default_repo_domain() : string {
 	if ( defined( 'FAIR_DEFAULT_REPO_DOMAIN' ) ) {
-		return FAIR_DEFAULT_REPO_DOMAIN;
+		return (string) \FAIR_DEFAULT_REPO_DOMAIN;
 	}
 
 	return 'api.aspirecloud.net';
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -706,7 +706,7 @@ function get_hashed_filename( $metadata ) : string {
 	}
 
 	// WP plugins must include the plugin filename.
-	if ( 'wp-plugin' === $metadata->type ?? '' ) {
+	if ( 'wp-plugin' === ( $metadata->type ?? '' ) ) {
 		return $slug . '/' . $file;
 	}
 
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -687,21 +687,30 @@ function get_banners( $banners ) : array {
  * @return string
  */
 function get_hashed_filename( $metadata ) : string {
-	$filename = $metadata->filename;
-	$type = str_replace( 'wp-', '', $metadata->type );
 	$did_hash = '-' . get_did_hash( $metadata->id );
 
-	list( $slug, $file ) = explode( '/', $filename, 2 );
-	if ( 'plugin' === $type ) {
-		if ( ! str_contains( $slug, $did_hash ) ) {
-			$slug .= $did_hash;
-		}
-		$filename = $slug . '/' . $file;
-	} else {
-		$filename = $slug . $did_hash;
+	// Use the slug from the filename, if present.
+	list( $slug, $file ) = array_pad( explode( '/', $metadata->filename ?? '', 2 ), 2, '' );
+	if ( '' === $slug ) {
+		$slug = $metadata->slug ?? '';
 	}
 
-	return $filename;
+	// Default to slug matching the plugin filename, if not specified.
+	if ( '' === $file ) {
+		$file = $slug . '.php';
+	}
+
+	// Append DID hash to slug if not already present.
+	if ( ! str_contains( $slug, $did_hash ) ) {
+		$slug .= $did_hash;
+	}
+
+	// WP plugins must include the plugin filename.
+	if ( 'wp-plugin' === $metadata->type ?? '' ) {
+		return $slug . '/' . $file;
+	}
+
+	return $slug;
 }
 
 /**
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -701,7 +701,7 @@ function get_hashed_filename( $metadata ) : string {
 	}
 
 	// Append DID hash to slug if not already present.
-	if ( ! str_contains( $slug, $did_hash ) ) {
+	if ( ! str_ends_with( $slug, $did_hash ) ) {
 		$slug .= $did_hash;
 	}
 
--- a/tests/phpunit/tests/Packages/GetHashedFilenameTest.php
+++ b/tests/phpunit/tests/Packages/GetHashedFilenameTest.php
@@ -108,6 +108,23 @@ class GetHashedFilenameTest extends WP_UnitTestCase {
 	}
 
 	/**
+	 * Test that a hash-like substring in the middle of the slug still gets the DID hash appended.
+	 */
+	public function test_should_append_hash_when_same_value_appears_in_middle_of_plugin_slug() {
+		$hash = get_did_hash( 'did:plc:example1234567890123456789' );
+		$slug = 'vendor-' . $hash . '-plugin';
+		$metadata = $this->create_metadata_document(
+			[
+				'filename' => $slug . '/' . $slug . '.php',
+				'slug' => $slug,
+				'type' => 'wp-plugin',
+			]
+		);
+
+		$this->assertSame( $slug . '-' . $hash . '/' . $slug . '.php', get_hashed_filename( $metadata ) );
+	}
+
+	/**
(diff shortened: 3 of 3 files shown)


inc/packages/namespace.php
```diff
@@ -687,21 +687,30 @@ function get_banners( $banners ) : array {
  * @return string
  */
 function get_hashed_filename( $metadata ) : string {
-	$filename = $metadata->filename;
-	$type = str_replace( 'wp-', '', $metadata->type );
 	$did_hash = '-' . get_did_hash( $metadata->id );
 
-	list( $slug, $file ) = explode( '/', $filename, 2 );
```
kasparsd, line 697: When `$metadata->filename` isn't specified (such as https://github.com/fairpm/fair-plugin/releases/latest/download/fair-metadata.json), the `explode()` failed with:

> Deprecated: explode(): Passing null to parameter #2 ($string) of type string is deprecated in /inc/packages/namespace.php on line 694

and the `$file` was undefined which caused:

> Warning: Undefined array key 1 in /inc/packages/namespace.php on line 694

inc/default-repo/namespace.php
```diff
@@ -26,7 +26,7 @@ function bootstrap() {
  */
 function get_default_repo_domain() : string {
 	if ( defined( 'FAIR_DEFAULT_REPO_DOMAIN' ) ) {
-		return FAIR_DEFAULT_REPO_DOMAIN;
+		return (string) \FAIR_DEFAULT_REPO_DOMAIN;
```
kasparsd, line 29: Ensure the output type is always correct and use the constant value from the global namespace since that is what `defined( 'FAIR_DEFAULT_REPO_DOMAIN' )` checks for.

inc/packages/namespace.php
```diff
@@ -687,21 +687,30 @@ function get_banners( $banners ) : array {
...
+		$slug = $metadata->slug ?? '';
+	}
+
+	// Default to slug matching the plugin filename, if not specified.
+	if ( '' === $file ) {
+		$file = $slug . '.php';
+	}
+
+	// Append DID hash to slug if not already present.
+	if ( ! str_contains( $slug, $did_hash ) ) {
```
costdev: Should use `str_ends_with()` to be safe here, since we're appending it when it's not there.

inc/packages/namespace.php
```diff
@@ -687,21 +687,30 @@ function get_banners( $banners ) : array {
...
+		$slug .= $did_hash;
+	}
+
+	// WP plugins must include the plugin filename.
+	if ( 'wp-plugin' === $metadata->type ?? '' ) {
```
```suggestion
	if ( 'wp-plugin' === ( $metadata->type ?? '' ) ) {
```
Need parentheses or it will not fall back to the empty string, or will throw a warning should the property be missing.

3v4l: https://3v4l.org/KYCPq

costdev (dismissed):

inc/packages/namespace.php
kasparsd: Ha, yeah. I wanted to do that but decided to keep it consistent with the current approach. Will update it.

kasparsd: Done!

costdev (commented):
Thanks for the changes @kasparsd! I won't get to test this or do a full review, but hopefully someone has availability and can help determine if any further tests would be good here.

kasparsd:
All the code review feedback has been addressed. Added a few additional phpunit test cases to account for the edge cases resolved.

inc/packages/namespace.php
afragen: I think the problem is that `$metadata->filename` is supposed to exist and that the `fair-metadata.json` was incorrectly created.

costdev:
@cdils I won't get to do a full review of this PR. If someone else if available to do a full review, that would be great!

Edit: It looks like @afragen started a review

inc/packages/namespace.php
kasparsd: Yeah, the general package spec doesn’t require it so I think it makes sense that we support that scenario as a fallback.

https://github.com/fairpm/fair-protocol/blob/83be5fa21fc1f72d5bcbf2bdd654f9e72274bc39/specification.md?plain=1#L197

Arguably, WP determines the bootstrap file dynamically based on header presence (and themes don’t even need it) so I think should probably refactor this eventually.

afragen: We did update the spec and it now requires it.

inc/packages/namespace.php
```diff
@@ -687,21 +687,30 @@ function get_banners( $banners ) : array {
...
+		$slug .= $did_hash;
+	}
+
+	// WP plugins must include the plugin filename.
+	if ( 'wp-plugin' === ( $metadata->type ?? '' ) ) {
+		return $slug . '/' . $file;
 	}
 
-	return $filename;
+	return $slug;
```
afragen, line 716: Also, not sure I understand why a function named `get_hashed_filename()` no longer returns `$filename`.

cdils:
@costdev sorry–I didn't mean to trigger a review from you as I saw your earlier comment. I was attempting to get rid of resolved conversations pertaining to the previous review. Blame it on my github skillz. :)

inc/packages/namespace.php
kasparsd: It returns the filename only for `wp-plugin` type and returns the slug in _all_ other cases. That is what this updated approach makes clear with a conditional return for that one edge case for WP plugins.

Feel free to push an update or I can update it if you can advise on the updated logic here.

costdev:
No worries haha!

inc/packages/namespace.php
kasparsd: @afragen I think the spec shouldn't require it. None of the WP packages have that concept (at least from the existing WP-org API perspective).

I created #500 to illustrate how WP core resolves the `plugin_file` dynamically during install and never gets it from the update API. Asking package publishers to include the filename is simply unnecessary.

In #500 I demonstrate how we can resolve all the necessary filename data from the canonical places in WP without having to assume or deal with mismatches on the filesystem (what `active_plugins` thinks) and what the metadata reports.

afragen:
Is there a test for the current state of the FAIR Connect where fair-plugin/plugin.php doesn’t list a filename in the metadata?

kasparsd:
@afragen Yes, this pull request adds those test cases, including this one with the missing `filename`:

https://github.com/kasparsd/fair-plugin/blob/a29caf53ec45d65666c92c8b78edaf41860228f6/tests/phpunit/tests/Packages/GetHashedFilenameTest.php#L34-L47

My proposal is to remove the `filename` requirement eventually since they're not really required as WP resolves the plugin file automatically by scanning all root-level PHP files for the `Plugin Name` header during install/activation. Created #500 to demonstrate that. But that's for a separate discussion.

cdils:
@kasparsd Can you update status on readiness to merge this PR? It looks like it needs to be rebased against `release_1.4.1` but not sure beyond that.

kasparsd:
@cdils This is ready for merge, and it appears to be targeting [fairpm:release_1.4.1](https://github.com/fairpm/fair-plugin/tree/release_1.4.1) already.

cdils (approved):

PR #497, opened 2026-05-16
Handle singleton artifact objects per FAIR spec

kasparsd:
Fixes #496.

This changeset fixes package handling when a FAIR repository returns `releases[].artifacts.package` as a single object instead of an array.

Per the FAIR core spec, the `artifacts` map allows each artifact type value to be either an object or a list of objects. The plugin was normalizing neither shape internally and some consumer paths assumed array access, which caused failures such as `Cannot use object of type stdClass as array` during install/update flows.

This patch normalizes release artifact values to arrays at parse time in `ReleaseDocument::from_data()`, so downstream code can consume a consistent internal shape regardless of how the repository serialized the artifact. Consumer paths that read package download URLs or inspect package artifacts were updated accordingly, and regression tests were added to cover both valid spec shapes.

**Spec references**

- FAIR Release Document: `artifacts` is a JSON object whose values “MUST be objects or lists of objects”
- FAIR Core compatibility implication: clients must accept both valid shapes for artifact values, including `artifacts.package`

**Testing**

Added parser regression tests for:
  - singleton artifact object -> normalized to array
  - artifact array -> preserved as-is

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -470,6 +470,9 @@ function pick_artifact_by_lang( array $artifacts, ?string $locale = null ) {
 
 		return $score;
 	};
+
+	// Minimal behavior only: missing lang is tolerated, but equal-score ordering
+	// and partial-locale fallback precedence are intentionally unspecified.
 	usort( $artifacts, function ( $a, $b ) use ( $score_artifact ) {
 		$a_score = $score_artifact( $a );
 		$b_score = $score_artifact( $b );
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -726,6 +726,7 @@ function get_package_data( $did ) {
 	$type = str_replace( 'wp-', '', $metadata->type );
 	$sections = (array) $metadata->sections;
 	$description = trim( $sections['description'] ?? '' );
+	$package_artifact = isset( $release->artifacts->package ) ? pick_artifact_by_lang( $release->artifacts->package ) : null;
 
 	$response = [
 		'name'              => $metadata->name,
@@ -747,8 +748,8 @@ function get_package_data( $did ) {
 		'new_version'       => $release->version,
 		'version'           => $release->version,
 		'remote_version'    => $release->version,
-		'package'           => $release->artifacts->package[0]->url,
-		'download_link'     => $release->artifacts->package[0]->url,
+		'package'           => $package_artifact->url ?? '',
+		'download_link'     => $package_artifact->url ?? '',
 		'tested'            => $required_versions['tested_to'] ?? '',
 		'external'          => 'xxx',
 		'last_updated'      => $metadata->last_updated ?? '',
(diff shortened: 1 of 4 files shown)


inc/packages/namespace.php
```diff
@@ -460,15 +460,19 @@ function pick_artifact_by_lang( array $artifacts, ?string $locale = null ) {
 	// Score artifacts based on match.
 	$score_artifact = function ( $artifact ) use ( $langs ) {
 		$score = 0;
+		$lang = strtolower( $artifact->lang ?? '' );
```
kasparsd, line 463: Account for missing `lang` property which is allowed per spec.

inc/packages/namespace.php
```diff
@@ -791,10 +796,7 @@ function cache_did_for_install( array $options ): array {
 		$did = array_find_key(
 			$releases,
 			function ( $release ) use ( $options ) {
-				if ( ! is_array( $release->artifacts->package ) ) {
-					return false;
-				}
-				$artifact = pick_artifact_by_lang( $release->artifacts->package );
+				$artifact = isset( $release->artifacts->package ) ? pick_artifact_by_lang( $release->artifacts->package ) : null;
```
afragen, line 799: Should this return `false` for the check below?

inc/packages/namespace.php
```diff
@@ -976,8 +978,12 @@ function maybe_add_accept_header( $args, $url ) : array {
 	}
 
 	foreach ( $releases as $release ) {
-		if ( $url === $release->artifacts->package[0]->url ) {
-			$content_type = $release->artifacts->package[0]->{'content-type'};
+		$artifact = array_find(
+			$release->artifacts->package ?? [],
```
costdev, line 982: This PR normalizes to an array of artifacts for a newly created `ReleaseDocument` via its `::from_data()` method.

However, since the Transient API is a persistent cache mechanism, is may be possible that an existing pre-1.5.0 `CACHE_RELEASE_PACKAGES` transient value could still have the non-normalized `package` and therefore this `array_find` call could run on an object.

The normalization may need to run against `$release->artifacts` here before the `array_find()` call, and it may be worth abstracting the artifact normalization to a new `static` method on the `ReleaseDocument` class to reduce code duplication.

inc/packages/namespace.php
costdev: Functionally speaking, `null` and `false` are both falsy so they'll both fail the truthy `$artifact` check below as expected.

Semantically speaking, this code is saying "The release artifact is..." (`object|null`), rather than "This has a release artifact" (`bool`), so `null` makes more sense to me when no release artifact exists.

Edit: `pick_artifact_by_lang()` returns `stdClass|null`, so this maintains a consistent fallback value type.

tests/phpunit/tests/Packages/PickArtifactByLangTest.php
```diff
@@ -0,0 +1,53 @@
...
+		$first_artifact = (object) [
+			'url' => 'https://example.com/first.zip',
+		];
+		$second_artifact = (object) [
+			'url' => 'https://example.com/second.zip',
+		];
+
+		$selected = pick_artifact_by_lang( [ $first_artifact, $second_artifact ], 'de-DE' );
+
+		$this->assertContains( $selected, [ $first_artifact, $second_artifact ], 'Artifacts without lang should still return a valid artifact.' );
```
costdev, line 51: Double-checking the reasoning why this checks that `$selected` is in an array rather than being a specific artifact. Is it because of inconsistent sorting across PHP versions or?

PR #499, opened 2026-05-18
pin workflow action versions to sha hash, tighten permissions

chuckadams:
Not sure if this should even be in the release branch, given the nature of it, but I'll let someone else decide that.  At any rate, they can't be tested until they land in `main`.

Rationale here: https://github.com/orgs/fairpm/discussions/226#discussioncomment-16965503

* Pins most actions to SHA hashes, except for a couple that stubbornly refuse to be pinned that way (that's happened to me on other projects, I couldn't tell you why)
* Adds permission restrictions wherever they're missing (most actions already had them).
* Removes some interpolations from `if:` clauses, since they're regarded as syntax errors, though in these cases GH does DWIM them anyway so the change is just cosmetic.
* Sets persist-credentials in the checkout action to false, except where the git CLI is used.

Several actions did get version upgrades as part of the pinning, but I don't think any of our actions are actually backward-incompatible.


joedolson (dismissed):

kasparsd:
Do we mind merging this in after #493 since they both touch the CI yaml files (to reduce conflicts).

Ipstenu (approved):

PR #500, opened 2026-05-18
Decouple package filename from package metadata

kasparsd:
Filename is a dynamic property that WP resolves during runtime. It shouldn't be part of package metadata.

1. During **an update** WP already knows the plugin bootstrap (basename) file from the option: https://github.com/WordPress/wordpress-develop/blob/5a96ff4d54a97955b02ac3c20f3ae7f21185232f/src/wp-includes/update.php#L570

2. And during **a fresh plugin install** it resolves it by scanning all PHP files in the root of newly installed plugin for the header: https://github.com/WordPress/wordpress-develop/blob/5a96ff4d54a97955b02ac3c20f3ae7f21185232f/src/wp-admin/includes/plugin-install.php#L469

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -731,8 +731,6 @@ function get_package_data( $did ) {
 		'author'            => $metadata->authors[0]->name,
 		'author_uri'        => $metadata->authors[0]->url,
 		'slug'              => $metadata->slug,
-		$type               => $filename,
-		'file'              => $filename,
 		'slug_didhash'      => $hashed_slug,
 		'url'               => $metadata->url ?? $metadata->slug,
 		'sections'          => $sections,
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -680,34 +680,24 @@ function get_banners( $banners ) : array {
 }
 
 /**
- * Get hashed file name from MetadataDocument.
+ * Get hashed slug from MetadataDocument.
  *
  * @param  MetadataDocument $metadata MetadataDocument.
  *
  * @return string
  */
-function get_hashed_filename( $metadata ) : string {
+function get_hashed_slug( $metadata ) : string {
 	$did_hash = '-' . get_did_hash( $metadata->id );
 
 	// Use the slug from the filename, if present.
-	list( $slug, $file ) = array_pad( explode( '/', $metadata->filename ?? '', 2 ), 2, '' );
+	list( $slug ) = array_pad( explode( '/', $metadata->filename ?? '', 2 ), 1, '' );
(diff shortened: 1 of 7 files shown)


inc/packages/wp-cli/compat/namespace.php
```diff
@@ -107,7 +101,37 @@ function handle_command( string $command, string $subcommand, array $args, array
...
 		case 'uninstall':
 		case 'update':
-			$args = array_merge( [ $command, $subcommand ], $hashed_items );
+			$packages = get_packages();
+			$plugin_basename_by_did = [];
+
+			// Get the plugin basenames for the DIDs.
+			foreach ( $dids as $did ) {
+				if ( ! empty( $packages['plugins'][ $did ] ) ) {
+					$plugin_basename_by_did[ $did ] = plugin_basename( $packages['plugins'][ $did ] );
```
kasparsd, line 110: Here we collect the plugin basenames from what WP core reports via `get_plugins()` and `wp_get_themes()` which has the correct filename data, instead of asking for that data to `Packages\get_hashed_filename( $metadata )` which:

1. doesn't know about the actual paths on the filesystem,
2. and could report a different filename than what is on the filesystem.

inc/packages/namespace.php
```diff
@@ -680,28 +680,27 @@ function get_banners( $banners ) : array {
...
+ * Get hashed slug from MetadataDocument.
  *
  * @param  MetadataDocument $metadata MetadataDocument.
  *
  * @return string
  */
-function get_hashed_filename( $metadata ) : string {
-	$filename = $metadata->filename;
-	$type = str_replace( 'wp-', '', $metadata->type );
+function get_hashed_slug( $metadata ) : string {
```
kasparsd, line 689: Never attempt to return the package filename since it can't be done reliably and can mismatch what is on the filesystem. This function now reports the desired slug/directory for this package.

inc/packages/namespace.php
```diff
@@ -722,21 +721,16 @@ function get_package_data( $did ) {
...
-	$sections = (array) $metadata->sections;
 	$description = trim( $sections['description'] ?? '' );
 
 	$response = [
 		'name'              => $metadata->name,
 		'author'            => $metadata->authors[0]->name,
 		'author_uri'        => $metadata->authors[0]->url,
 		'slug'              => $metadata->slug,
-		'slug_didhash'      => $metadata->slug . '-' . get_did_hash( $did ),
-		$type               => $filename,
```
kasparsd, line 736: This is injected during `update_site_transient()` instead when we actually know the current filename on the filesystem which is what WP core uses for the lifecycle of the update.

PR #511, opened 2026-06-10
carpet bomb fair-plugin with tests

chuckadams:
Tests, tests, and more tests.  Unit tests, integration tests, http tests, end-to-end tests, all for one low low price.  99.9% AI generated, now in the human review stage.

Only one source file changed, and that's two trivial changes in inc/packages/namespace.php: one to fix a typeerror crash when `reset([])` returns `false`, and the other to allow customizing the plc.directory url which allows us to mock it in integration tests.  That's also being backported into 1.4.1, so by the time this lands it should be zero source files changed.

There's also a couple github workflows, one to run coverage and upload it as an artifact, another to run integration tests which is workflow_dispatch only until I know exactly how slow it is on github runners.   The existing release workflow now builds with PHP 8.0 instead of 8.2 so that it doesn't bundle a vendor/ with deps that are only for 8.2+.  There shouldn't be any surprises, since we test against 8.0 already, including doing a `composer update` in each test matrix cell.

Lots of ai-generated planning docs in ai/plans.  I can nuke those if we want, I don't usually keep completed plans in git, just figured they might be edifying.  The ai/ directory won't ship in release zips.


kasparsd:
Would it be possible to split this out into separate pull requests for clarity:

1. one that extends the phpunit test suite
2. one that adds phpunit coverage reporting
3. one that adds the e2e tests

I feel like there are some ways to simplify the e2e setup by relying on existing WP packages and helpers but it wouldn't be fair to the phpunit suite work if we started discussing it here 😂

chuckadams:
Possible for sure, but in the end it's all tests, and they're quite well separated by function into different directory trees under test/.    I'm not in a massive rush to merge this, and and we can make plenty of changes as needed.

As for e2e tests in particular, I'm rather attached to playwright, and would need a lot of convincing evidence to switch to something else.  Ultimately it's about what harness is controlling Chromium/Safari/FF, because anything else is not actually a browser test.

PR #512, opened 2026-06-10
Trivial tweaks to inc/packages/namespace.php

chuckadams:
First one makes the plc.directory url configurable.  This is already used in the `test_the_things` branch to set up a mock PLC server so we don't have to mock the http transport itself.

Second one fixes a crash if someone passes an empty versions list to `pick_release`, since `reset([])` returns false (btw  have I mentioned what a fractal of bad design `reset` is?)


afragen (approved):

kasparsd (commented):
Looks good!

PR #526, opened 2026-08-27
Skip package directory rename (to DID has suffix) on update

### LINKED ISSUE #502: 🐞 Bug Report: Upgrades from 1.4.0 disable the FAIR Connect plugin because of directory rename
## Detailed Description

Updates handled by the DID update path in 1.4.0 now run the plugin directory rename sequence which adds the DID suffix to the plugin slug. So `plugins/fair-plugin` becomes `plugins/fair-plugin-{hash}`.

When this happens, the plugin gets disabled because the option holding the active plugins no longer points to the updated FAIR connect plugin path.

## Expected Behavior

FAIR Connect should keep itself activate after upgrades.

## Steps to Reproduce

1. Install 1.4.0 of the plugin (using ZIP downloaded from GitHub or the project website).
2. Downgrade it to 1.3.9 by updating the `Version` string in the main plugin file.
3. Go to WP updates and click on update.
4. After update the plugin is no longer active.

Image

kasparsd:
Fixes #502.

Updates must keep the installed basename stable: core derives the destination, the old package deletion, and the `active_plugins` entry from `hook_extra['plugin']`. Renaming the source to `slug-didhash` during updates invalidated the active entry and deactivated the plugin.

Bulk updates pass no `action` key in `hook_extra`, so bail when it is missing too.

### DIFF
--- a/inc/packages/namespace.php
+++ b/inc/packages/namespace.php
@@ -833,6 +833,11 @@ function delete_cached_did_for_install(): void {
  *
  * This is commonly required for packages from Git hosts.
  *
+ * Only applies to installs. Updates must keep the installed package basename
+ * stable: core derives the destination, the old package deletion, and the
+ * active_plugins entry from hook_extra['plugin'], so renaming the source
+ * during an update would orphan the active plugin entry and deactivate it.
+ *
  * @param string|WP_Error $source Path of $source, or a WP_Error object.
  * @param string $remote_source Path of $remote_source.
  * @param WP_Upgrader $upgrader An Upgrader object.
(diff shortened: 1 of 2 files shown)


inc/packages/namespace.php
```diff
@@ -850,8 +855,9 @@ function maybe_rename_on_package_download( $source, string $remote_source, WP_Up
 		return $source;
 	}
 
-	// Exit early if installing.
```
kasparsd, line 853: @afragen you might have context for this -- do you remember the reasoning for applying the directory rename on non-installs? 

I believe the desired logic is to keep the slug unchanged at all times _except for the initial install_ when there is no history of the plugin on the site, right?

